PR #1, opened 2016-06-07
Control nodes

cladmi:
Add support for multiple gateways


aabadie:
Travis fails because there's no travis configuration file in this branch and `the skip if no .travis.yml` option was not checked in travis config.

aabadie:
Rebase and merge as you wish (also check that tests still pass before)

PR #2, opened 2016-06-07
Samr21 xpro

cladmi:
Add samr21-xpro support


cladmi:
- [x] Currently 'leds_on' and 'leds_off' commands are broken on the firmware.

But as there is no 'leds_consumption' feature, they are not tested.
- [x] I should improve autotests to make this visible, and then tests leds commands.
- [ ] Also, I may need to add another entry to tell the number of leds to be able to test consumption.

PR #4, opened 2016-10-18
Add Zigduino board previous fork

schrein:
Add support for open node Zigduino. More info on http://wiki.logos-electro.com/zigduino-r2-manual.


cladmi:
I set up a custom zigduino on CI nodes and added this branch to integration tests as: 

> https://github.com/iot-lab/iot-lab-gateway/tree/screin/pr/custom-zigduino.

However, I saw errors on autotest cleanup test when running manually:

> https://gist.github.com/cladmi/398b220cd848554d0648f7c367ada783

I think it's because the node is not ready directly after connecting to the serial port.
It should also break other tests later.
Will be for next week though.

tests_utils/integration_fabfile.py
```diff
@@ -37,7 +37,7 @@
 # Default to all targets
 if not env.hosts:
     env.hosts = ['leonardo-00-ci', 'm3-00-ci', 'a8-00-ci', 'fox-00-ci']
-    env.hosts += ['samr21-00-ci']
+    env.hosts += ['samr21-00-ci', 'zigduino-00-ci']
```
rienafairefr, line 40: This shouldnt be needed anymore with the new fabric ci development on install lib,

fsaintma: yes but I didn't deploy the new ci management with intsall-lib and ci.inria.fr
so it will be not needed :)

schrein:
Currently rebase to latest master and now need more tests before merging. See work in progresse here : https://github.com/schrein/iot-lab-gateway/tree/zigduino .

schrein:
Work in progress here https://github.com/iot-lab/iot-lab-gateway/pull/28

PR #7, opened 2017-07-12
Dockerize the iot-lab-gateway application

c-roussel:
The purpose of dockerizing the gateway's REST application was to give an accessible development environment to actual and future developers.

Using the Docker-run.sh script, a user who built (Docker-build.sh) or has the gateway's docker image can either start a container with its own Gateway_code directory mounted or not. This allows developers to make profit from a reloader option added to the REST application.


rienafairefr:
I've added getting liboml2 source from https://github.com/mytestbed/oml, apparently the mytestbed.net website is down. Checking out release/2.11 branch, autogen then build. I've run the Docker-build.sh without issues.

Docker-run.sh
```diff
@@ -0,0 +1,71 @@
...
+  fi
+fi
+
+cat<<EOF
+Used config :
+  BOARD_TYPE=$BOARD_TYPE, HOSTNAME=$HOSTNAME, CONTROL_NODE_TYPE=$CONTROL_NODE_TYPE
+  Montage : $OPT_DIR
+EOF
+
+docker run \
```
rienafairefr, line 59: I think --rm and -d conflict in docker 1.12, this can only be run in docker 1.13 

(see https://docs.docker.com/release-notes/docker-engine/#1130-2017-01-18
https://github.com/docker/docker/pull/20848)

aabadie:
Some documentation would be great as well.

vandaele:
Superseded by #13

PR #9, opened 2017-08-29
Add b-l072z-lrwan1 board

aabadie:
This adds the support for ST [STM32L0 Discovery kit LoRa](http://www.st.com/en/evaluation-tools/b-l072z-lrwan1.html) open node.

This boards requires OpenOCD 0.10 so the actual port won't work before we modify the gateway code to support both openocd 0.9 and openocd 0.10.


aabadie:
Will reopen with a branch from my fork

rienafairefr:
if it's safe to delete the l0-lrwan1 branch on this repo don't hesitate to do it ^^

aabadie:
Thanks for the reminder. Done

PR #10, opened 2017-09-21
Add Zolertia Firefly board

vandaele:
Adds support for the (Zolertia Firefly)[https://github.com/Zolertia/Resources/wiki/Firefly] board.

This boards requires the cc2538 Boot Serial Loader to flash firmware. Support in the gateway code and install instructions have been added, using https://github.com/JelmerT/cc2538-bsl.


gateway_code/utils/CC2538.py
```diff
@@ -38,7 +38,8 @@ class CC2538(object):
     DEBUG = ('')
 
     @property
-    def add_command(self):
+    def addr_command(self):
+        #TODO Do it a better way with python
```
vandaele: There's a TODO here.
Could be better to replace this long and complex shell command above by python code.

bin/rules.d/firefly.rules
```diff
@@ -0,0 +1,9 @@
+# open nodes firefly
+# previous work by Cedric
+# KERNEL=="ttyUSB[0-9]*" , SUBSYSTEMS=="usb", ATTRS{idVendor}=="10c4" , ATTRS{idProduct}=="ea60" , ATTRS{product}=="Zolertia Firefly platform" , MODE="0664", GROUP="dialout", SYMLINK+="ttyON_firefly"
```
rienafairefr: This was taken from http://zolertia.sourceforge.net/wiki/index.php/Mainpage:Apps

setup.py
```diff
@@ -73,7 +73,7 @@ def get_version(package):
 SCRIPTS = glob('bin/scripts/*')
 
 INSTALL_REQUIRES = ['argparse', 'bottle', 'paste', 'pyserial']
-INSTALL_REQUIRES += ['pyelftools']
+INSTALL_REQUIRES += ['pyelftools', 'intelhex', 'python-magic']
```
aabadie: Those new packages don't seem to be used in the changes. Are they really required ?

rienafairefr: it's due to the cc2538 dependency. 
python-magic is not fully needed, I think I'll just write temporary firmwares to flash with a .hex extension and not add python-magic.

aabadie: you mean the cc2538-bsl.py tool ? this is an external tool so no need to add this dependency to the gateway code. I think it's enough to just have it as a dependency in the docker container and a note in the readme (like it's already done)

rienafairefr: You're right

rienafairefr:
OK this has been tested in local, code_check. Normally the latest `master` yocto images already have the cc2538-bsl added so we could merge that and plug a node in ci

aabadie:
Let's go then !

PR #11, opened 2017-09-25
metaclass based plugin system for open/control nodes

rienafairefr:
Simplify the way custom open/control nodes are added to the gateway code, no need for a class named Node*, a file named cn_ or on_, everything is verified through abc.abstract*, so we can't create a class or object that is missing some methods.

open nodes inherit from OpenNode, control nodes inherit from ControlNode, easy :-)


gateway_code/nodes.py
```diff
@@ -0,0 +1,170 @@
...
+    def __init__(cls, name, bases, class_dict):
+        type.__init__(cls, name, bases, class_dict)
+        if not hasattr(cls, '__registry__'):
+            cls.__registry__ = {}
+        else:
+            cls.__registry__[cls.TYPE] = cls
+
+
+class MetaNode(MetaNodeBase, abc.ABCMeta):
+    """ Combining metaclasses, control or open Nodes have a contract verified on instantiation """
```
aabadie: pep8: line is too long

gateway_code/control_nodes/tests/control_node_test.py
```diff
@@ -0,0 +1,17 @@
...
+
+from gateway_code.nodes import all_control_nodes_types, control_node_class
+
+
+class TestsControlNodesImplementations(unittest.TestCase):
+    """Test loading implemented control nodes implementations."""
+
+    @staticmethod
+    def test_nodes_classes():
+        """Test loading all implemented open nodes implementation."""
```
aabadie: It should be control nodes instead of open nodes, right ?

gateway_code/open_nodes/tests/open_nodes_test.py
```diff
@@ -21,43 +21,79 @@
 """open_nodes package tests."""
 
 import unittest
+
 import mock
 
-from .. import node_class, all_nodes_types
-from ..node_m3 import NodeM3
-from ..node_a8 import NodeA8
+from gateway_code.nodes import open_node_class, all_open_nodes_types, OpenNode, all_control_nodes_types, \
```
aabadie: pep8: line is too long

You can use multi line with imports inside parenthesis:
```python
from gateway_code.nodes import (open_node_class, all_open_nodes_types
                                OpenNode, all_control_nodes_types, control_node_class)
```

gateway_code/utils/tests/openocd_test.py
```diff
@@ -29,26 +29,27 @@
 
 import time
 import unittest
+
```
aabadie: you don't need this empty line

aabadie (changes requested):
Looks good and nice idea in general. There are some pep8 though and a few minor things.

rienafairefr:
OK I've modified the PR with reviews, thanks @aabadie

rienafairefr:
OK I've reorganized the commit history, made changes from reviews, what do you guys think

DEVELOPER.md
```diff
@@ -63,7 +63,7 @@ This section present the module architecture.
...
+Between two plugs, the name of the device as detected by the embedded linux running on the gateway can change: for example, an unique device can be succesively detected as ttyUSB0 and then as ttyUSB1. As it is essential to have a fixed name for your device, you have to write a udev rules specific for your device. To do such a thing, you must create a file named `your_node.rules` (with `your_node` the name of your node). Using the command `udevadm`, you have to retrieve some information to identify your node such as the serial id, the vendor id or the product id. You must set a name for the device and set the right group (dialout) and the right mode (664). It's required to name the device `ttyON_NODENAME`. Below the example of the udev rule for the M3 node:
```
aabadie: I had a hard time finding what was changed in this line (because it's a whole single line!). So while you are at it, you could split it to keep the 80 characters max width

DEVELOPER.md
```diff
@@ -137,7 +148,7 @@ a better idea of what the application expects.
 
 When one of the function described above terminates with a success, the
 return value must be 0. When the return value is different from 0, the
-application know that something gone wrong. Except status, all the method
+application knows that something went wrong. Except status, all the method
```
aabadie: all the method**s** ?

DEVELOPER.md
```diff
@@ -147,7 +158,8 @@ with the serial port of the open node. To do this, you will have to use
 a programmer software and write code to interact with it. `Openocd` and
 `Avrdude` are already available but maybe you will need to use another
 programmer and implement your own python file in the `gateway_code/utils/`
-folder.
+folder. If you need a specific kernel driver to be available, don't
```
aabadie: `If you need a specific kernel driver, ask the IoT-LAB admin team at admin@iot-lab.info`

That might become obsolete if one day user kernel/images are allowed on the testbed.

DEVELOPER.md
```diff
@@ -203,7 +215,7 @@ When developing your implementation, you should verify that all unit tests pass
 
 ### Integration tests
```
aabadie: Out of scope of this PR but maybe some notes on how to test using the container image could be added here ?

docker-gateway-rest-server
```diff
@@ -29,8 +29,8 @@ from gateway_code.config import control_node_class
 
 
 def manual_udev(env, clazz):
-    if env in os.environ and hasattr(clazz, 'TTY'):
-        cmd = 'ln -nfs %s %s' % (os.environ[env], clazz.TTY)
+    if env in os.environ and hasattr(clazz, 'tty'):
```
aabadie: This was something I found not very nice in #13. Why not rename `clazz` in `cls` ? or `node_cls` ?

gateway_code/control_nodes/cn_iotlab/__init__.py
```diff
@@ -34,11 +35,10 @@
 LOGGER = logging.getLogger('gateway_code')
 
 
-class ControlNodeIotlab(object):
+class ControlNodeIotlab(ControlNode):
     """ Control Node implemenation """
```
aabadie: s/implemenation/implementation/ (typo not related to your changes)

gateway_code/control_nodes/tests/control_node_test.py
```diff
@@ -0,0 +1,38 @@
...
+class TestsControlNodesImplementations(unittest.TestCase):
+    """Test loading implemented open nodes implementation."""
+
+    @staticmethod
+    def test_nodes_classes():
+        """Test loading all implemented control nodes implementation."""
+        for node in all_control_nodes_types():
+            # No exception
+            print(node)
+            cls = control_node_class(node)
```
aabadie: node_cls ?

gateway_code/nodes.py
```diff
@@ -0,0 +1,192 @@
...
+    def teardown(self):
+        """ Cleanup the open node """
+        pass
+
+    @abc.abstractmethod
+    def status(self):
+        """ Status of the open node """
+        pass
+
+    @property
```
aabadie: I think this method could be shared in the NodeBase class ?

gateway_code/open_nodes/tests/open_nodes_test.py
```diff
@@ -18,37 +18,30 @@
...
 
+from __future__ import print_function
 import unittest
-import mock
 
-from .. import node_class, all_nodes_types
-from ..node_m3 import NodeM3
-from ..node_a8 import NodeA8
+from gateway_code.nodes import open_node_class, all_open_nodes_types
+from gateway_code.open_nodes.node_m3 import M3
```
aabadie: Why not calling them iotlab_m3 and iotlab_a8 ?

gateway_code/tests/gateway_logging_test.py
```diff
@@ -23,13 +23,12 @@
 # pylint: disable=missing-docstring
 # pylint: disable=too-many-public-methods
 
-import os
 import tempfile
 import unittest
 import logging
+from StringIO import StringIO
```
aabadie: This seems unrelated but should be fixed after #18 I guess

aabadie (changes requested):
Another set of comments. But the general work is very nice.

I would prefer that other pending PRs are merged first and the CI (on Jenkins) passes again before merging this one.

DEVELOPER.md
rienafairefr: Yes, we'll update that doc when that's the case :-)

gateway_code/control_nodes/cn_iotlab/__init__.py
rienafairefr: Yes, I believe it's fixed somewhere else,I removed unrelated clutter from this PR

gateway_code/nodes.py
rienafairefr: OK, I've added a NodeBase class with the status method that is common between control and open nodes in it, that should remove that code duplication

gateway_code/open_nodes/tests/open_nodes_test.py
rienafairefr: True, that might make it clear that these are the reference implementations making it clear the difference with custom nodes

rienafairefr: I'll keep the node type attribute TYPE to `'m3'` and `'a8'` though, it's that way in install lib and other places

rienafairefr: It's gonna be IotlabM3 and IotlabA8 classes should be CamelCase

gateway_code/tests/gateway_logging_test.py
rienafairefr: Yes, I've tested locally a merge and it seems to not create conflict

rienafairefr:
I've cleaned up the code, we'll probably talk about this PR more next meeting

rienafairefr:
Just rebased and re-tested, integration tests (only open_nodes_test.py) are passing for all supported gateway open node types
```
{
 "Ok": [
  "a8,17", 
  "fox,1", 
  "leonardo,1", 
  "m3,17", 
  "samr21,1", 
  "zigduino,1"
 ]
{
 "Ok": [
  "arduino-zero,1", 
  "microbit,1", 
  "st-lrwan1,2"
 ]
}
}
```
Shouldn't this be merged already, it's been 10 months

gateway_code/nodes.py
```diff
@@ -0,0 +1,184 @@
...
+            cls.__registry__[cls.TYPE] = cls
+
+
+class MetaNode(MetaNodeBase, abc.ABCMeta):
+    """ Combining metaclasses,
+    control or open Nodes have a contract verified on instantiation """
+    pass
+
+
+class NodeBase(object):  # pylint: disable = too-few-public-methods
```
aabadie: I don't understand why you need 3 extra levels of abstract class here. Why not just do this:
```python
class NodeBase(metaclass=ABCMeta):

    def __init__(self):
        type.__init__(cls, name, bases, class_dict)
        if not hasattr(cls, '__registry__'):
            cls.__registry__ = {}
        else:
            cls.__registry__[cls.TYPE] = cls

    @abc.abstractmethod
    def status(self):
        pass

class ControlNode(NodeBase):
    TYPE = 'undefined'

# or maybe ? (not sure about that)

class ControlNode(metaclass=NodeBase):
    TYPE = 'undefined'
```

rienafairefr: It's because the Node metaclass need to inherit from both ABCMeta and the 'registry' metaclass, to have both behaviors. MetaNode is a combining metaclass, which is the metaclass for the ControlNode and OpenNode classes. 

Essentially, when seeing `class MyNode(OpenNode)`, Python looks for a metaclass in OpenNode, finds that it's MetaNode, and adds *both* a `__registry__` attribute to the MyNode class, and the way to check that the MyNode subclass implements the abstract methods defined in OpenNode.

i'll add some unit tests to demonstrate that

We could remove the NodeBase at the cost of a repeat of the `status` method though

rienafairefr:
I've pushed some more tests. If you want to experiment with metaclasses modification, run them with
`POSARGS="-e test -- --no-cov gateway_code/open_nodes/tests/open_nodes_test.py" make test`

The code in your comment [above](https://github.com/iot-lab/iot-lab-gateway/pull/11#discussion_r200400232) does not work or pass the tests  ;-)

The metaclass dance is necessary because we insist that open-node-support-designers use class-level attributes to define node properties. attributes on instances, higher level is classes. attributes on classes, higher level is metaclasses.

aabadie (changes requested):
I'd like to merge this PR soon but before I would like some changes:
- if possible, remove the MetaNode class and make NodeBase inherit from MetaNodeBase and abc.ABCMeta directly
- rename ControlNode to ControlNodeBase and OpenNode to OpenNodeBase
- revert the changes NodeXX to XX. This increases the size of the PR and is useless
- squash your commits

aabadie:
> The code in your comment above does not work or pass the tests ;-)

True, I was confusing with the way abc is used in another project.

gateway_code/nodes.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def all_open_nodes_types():
+    """Returns all the open nodes classes"""
+    return OpenNode.__registry__.keys()
+
+
+def all_control_nodes_types():
+    """Returns all the control nodes classes"""
+    return ControlNode.__registry__.keys()
```
aabadie: Just wondering: what happens with nodes that inherit from other nodes (example: ControlNodeIotlabM3 inherits ControlNodeIotlab which itself inherit ControlNode) ?
Does the registry key list contains both or only the direct child (ControlNodeIotlab)?

rienafairefr: inheritance work intuitively I think, if the thing after the `class` keyword defines a `TYPE` attribute, then that type will be used to store in the registry. The only potential trap is a subclass without an explicit TYPE attribute:

```
class BaseOpenNode(OpenNode):
   TYPE = 'base'

class TypeMissingDerivedOpenNode(BaseOpenNode):
   pass

assert open_node_class("base_open_node") == TypeMissingDerivedOpenNode
```


In the case of ControlNodeIotlab, the registry is working as intended (existing unit tests in gateway_code.control_nodes.tests)
```
iotlab
<class 'gateway_code.control_nodes.cn_iotlab.ControlNodeIotlab'>
iotlabm3
<class 'cn_iotlabm3.ControlNodeIotlabm3'>
no
<class 'cn_no.ControlNodeNo'>
```

aabadie: > The only potential trap is a subclass without an explicit TYPE attribute

This is exactly what I'm wondering. I'm thinking of factorizing some open nodes code like for example st-lrwan1 and st-iotnode which are 2 St-Link based open-nodes, we could make them derive from a `NodeStlink(OpenNode)` that would at least provide TTY, BAUDRATE that are common. The subclass won't have to explicitly provide them and IIUC we are in the trap you mention here.

rienafairefr: if parent class has no TYPE attribute then it might work though :-) I'm adding a unit test to show what I mean

rienafairefr: look at [test_parent_open_node_inheritance](https://github.com/iot-lab/iot-lab-gateway/pull/11/files#diff-4d189692dbb80bb871effc05f7bd5899R114) for an example of board classes inheriting from a base, the base is not visible in the registry, and the derived classes are valid open nodes, with the TTY and BAUDRATE inherited

aabadie: Ok, thanks! I'm convinced now :)

aabadie:
@rienafairefr, it would be great if you could address the other things I reported in [this comment](https://github.com/iot-lab/iot-lab-gateway/pull/11#pullrequestreview-144452129). After that, I'll merge.

rienafairefr:
I think it's better now (still needs to fix some lint though) To sum up:

* if possible, remove the MetaNode class and make NodeBase inherit from MetaNodeBase and abc.ABCMeta directly

I've removed the rather useless NodeBase (only needed to DRY the status method). MetaNode is the base metaclass, can't be removed, it's combining the registry metaclass and the abc.ABCMeta. ControlNode and OpenNode need it as its metaclass. If NodeBase could "pass on" its metaclass to it's child classes it would be ok, but it's not the way the class system works in Python.
I've enhanced it by inheriting from abc.ABCMeta directly (see 
5c64e61 ), seems to work and simplifies the class hierarchy 

* rename ControlNode to ControlNodeBase and OpenNode to OpenNodeBase
Any benefit why ? 

* revert the changes NodeXX to XX. This increases the size of the PR and is useless
Done

* squash your commits
Done, down to 8 commits at last count

aabadie:
> Any benefit why ?

Semantically, it means that this kind of class can't be instantiated directly and can only used using inheritance in subclasses.

rienafairefr:
OK I see. will rename OpenNode to OpenNodeBase and ControlNode to ControlNodeBase :+1:

rienafairefr:
OK tests are passing, good to merge for me

gateway_code/open_nodes/tests/open_nodes_test.py
```diff
@@ -18,46 +18,142 @@
 #
 # The fact that you are presently reading this means that you have had
 # knowledge of the CeCILL license and that you accept its terms.
-"""open_nodes package tests."""
+""" open nodes plugins tests """
 
-import unittest
-import mock
+# pylint: disable=missing-docstring
```
aabadie: Maybe you could really fix this ?

gateway_code/open_nodes/tests/open_nodes_test.py
```diff
@@ -18,46 +18,142 @@
 #
 # The fact that you are presently reading this means that you have had
 # knowledge of the CeCILL license and that you accept its terms.
-"""open_nodes package tests."""
+""" open nodes plugins tests """
```
aabadie: I think it was ok the way it was before

aabadie (approved):
OK, let's merge this one :beers:

aabadie:
Thanks @rienafairefr !

rienafairefr:
:v:

PR #12, opened 2017-10-10
add travis integration

vandaele:


.travis.yml
```diff
@@ -0,0 +1,12 @@
...
+
+services:
+  - docker
+
+before_install:
+  - docker build -t iot-lab-gateway .
+  - docker build -t iot-lab-gateway-test tests
+
+script:
+  - docker run -v gateway_code:/home/iot-lab-gateway/gateway_code -v $CWD/tests_utils/cfg_dir:/var/local/config -e BOARD_TYPE=m3 -e CONTROL_NODE_TYPE=no -e HOSTNAME=m3-00 -e IOTLAB_GATEWAY_NO_INTEGRATION_TESTS --privileged iot-lab-gateway-test
```
aabadie: Do you think this line could be split (<80 characters long) ?

rienafairefr: It'd be great. But I havent found a way to do that in a travis yml, yet

gateway_code/control_nodes/cn_no/__init__.py
```diff
@@ -57,9 +57,9 @@ def setup():
         return 0
 
     @staticmethod
-    @logger_call("Control node: Flash")
+    @logger_call("Control node : Flash")
```
aabadie: The space is not required (in english, there's no space before `:`)

setup.py
```diff
@@ -187,7 +186,7 @@ def run(self):
 
       scripts=SCRIPTS,
       include_package_data=True,
-      package_data={'static': ['static/*']},
+      package_data={'static': ['static/*'],'control_node_serial_interface':'control_node_serial/control_node_serial_interface'},
```
aabadie: should be split (pep8)

gateway_code/control_nodes/cn_no/__init__.py
```diff
@@ -57,9 +57,9 @@ def setup():
...
     def flash(*args, **kwargs):
-        """Setup control node."""
+        """ Flash ControlNode """
```
aabadie: Left over from #16 ;) Let's keep the change but remove the spaces before/after,  add a period and replace ControlNode with the previous version (in short, just replace `Setup` with `Flash`) !

.travis.yml
fambon: how about sticking this in a separate docker-test script then, or in a simple pytest ?
and it would give a usage example in the source tree :)

aabadie (commented):
Just minor things. Otherwise Travis is green, nice !

.travis.yml
rienafairefr:
```
script:
  - >
    docker run -v gateway_code:/home/iot-lab-gateway/gateway_code
    -v $CWD/tests_utils/cfg_dir:/var/local/config
    -e BOARD_TYPE=m3
    -e CONTROL_NODE_TYPE=no
    -e HOSTNAME=m3-00 
    -e IOTLAB_GATEWAY_NO_INTEGRATION_TESTS
    --privileged iot-lab-gateway-test
```
might work, if travis doesnt like it I'll put that in a tests_utils/docker-test

gateway_code/control_nodes/cn_no/__init__.py
rienafairefr: ok, fixed in last rebase, I think

setup.py
rienafairefr: ok, fixed in last rebase

gateway_code/control_nodes/cn_no/__init__.py
rienafairefr: damn, this makes a merge conflict with master, not sure why

aabadie:
This is because your branch contains commits from #16 that are now in master (but with a different hash). The way to fix that is to rebase in interactive mode and remove all commits that are not only related to this PR

gateway_code/control_nodes/cn_no/__init__.py
rienafairefr: OK, I'm dumb, I rebased on an outdated master

gateway_code/control_nodes/cn_no/__init__.py
```diff
@@ -59,7 +59,7 @@ def setup():
     @staticmethod
     @logger_call("Control node: Flash")
     def flash(*args, **kwargs):
-        """Setup control node."""
+        """Flash ControlNode"""
```
aabadie: Still "ControlNode" => "control node."

aabadie:
There is still the problem of running Travis build on all branches of the upstream repo: that's why we have 2 check entries in the PR.
One cannot filter the branch names (e.g master) to skip other branch name. This forces us to use branch from personal fork for PR.

.travis.yml
```diff
@@ -0,0 +1,19 @@
+language: generic
+sudo: required
+
+services:
+  - docker
+
+before_install:
+  - docker build -t iot-lab-gateway .
```
aabadie: If this line really needed ?
the iot-lab-gateway-test could be based on the released image on docker hub ?

aabadie: If it's not needed the travis build should be a little bit faster

rienafairefr: Don't think it would work. We want to test the current committed state, not the state of the released image. 

Right now: If test fails, then either the image or the gateway_code is bad, and both are at the same commit
If you test using a released image: If test fails, then either the image or the gateway_code is bad, and they are not at the same commit

it'd be great if we could cache the docker layers from dependencies (that's what's very long), and I think there are features in travis to maybe do that https://docs.travis-ci.com/user/caching/

.travis.yml
```diff
@@ -0,0 +1,32 @@
...
+    docker run -v gateway_code:/home/iot-lab-gateway/gateway_code
+    -v $CWD/tests_utils/cfg_dir:/var/local/config
+    -e BOARD_TYPE=m3
+    -e CONTROL_NODE_TYPE=no
+    -e HOSTNAME=m3-00 
+    -e IOTLAB_GATEWAY_NO_INTEGRATION_TESTS
+    --privileged iot-lab-gateway-test
+
+
+after_success:
```
aabadie: Should this be part of this PR ? I don't think so. Please wait after we are done with setting up travis. Otherwise, it will become a mess.

rienafairefr: ok. I think if the latest build is green this could be merged and close

aabadie: Yes, that's what I wanted to do (just testing locally with arduino zero right now)

setup.py
```diff
@@ -187,7 +186,8 @@ def run(self):
 
       scripts=SCRIPTS,
       include_package_data=True,
-      package_data={'static': ['static/*']},
+      package_data={'static': ['static/*'],
+                    'control_node_serial_interface':'control_node_serial/control_node_serial_interface'},
```
aabadie: this is a pep8: line exceeds 80 characters (I'm wondering why the CI still pass

.travis.yml
```diff
@@ -0,0 +1,20 @@
...
+  - docker
+
+before_install:
+  - docker pull fitiotlab/iot-lab-gateway:latest
+  - docker build -t iot-lab-gateway .
+  - docker build -t iot-lab-gateway-test tests
+
+script:
+  - >
+    docker run -v gateway_code:/home/iot-lab-gateway/gateway_code
```
aabadie, line 14: I tried the commands locally and this one fails. It looks like my changes are not taken into account. But maybe I missed something

setup.py
rienafairefr: code_check is not run in CI, that's why :-)

.travis.yml
rienafairefr: what's the error ?
you've built both images before the docker run ?
```
docker build -t iot-lab-gateway .
docker build -t iot-lab-gateway-test tests
```

setup.py
```diff
@@ -176,6 +175,10 @@ def run(self):
         execute(self, post_install, [self])
 
 
+control_node_serial_data = 'control_node_serial/control_node_serial_interface'
```
aabadie: The variable name is not very related to its content (more related to serial interface). By the way, you don't really need an extra variable here, just split the line below

.travis.yml
aabadie: yes, I have errors (always the same, even if I change the setup.cfg) related to coverage:
```
nose.plugins.cover: WARNING: Failed to generate HTML report: No source for code: '/home/iot-lab-gateway/gateway_code/control_node/__init__.py'.
Aborting report output, consider using -i.
Traceback (most recent call last):
  File "setup.py", line 199, in <module>
    install_requires=INSTALL_REQUIRES)
  File "/home/iot-lab-gateway/.tox/check/local/lib/python2.7/site-packages/setuptools/__init__.py", line 129, in setup
    return distutils.core.setup(**attrs)
  File "/usr/lib/python2.7/distutils/core.py", line 151, in setup

... (long traceback)

    raise NoSource(exc_msg)
coverage.misc.NoSource: No source for code: '/home/iot-lab-gateway/gateway_code/control_node/__init__.py'.
Aborting report output, consider using -i.
ERROR: InvocationError: '/home/iot-lab-gateway/.tox/check/bin/python setup.py nosetests'
___________________________________ summary ____________________________________
  copying: commands succeeded
ERROR:   check: commands failed
```

aabadie: I'm running: 
```
docker run -v gateway_code:/home/iot-lab-gateway/gateway_code -v $CWD/tests_utils/cfg_dir:/var/local/config -e BOARD_TYPE=m3 -e CONTROL_NODE_TYPE=no -e HOSTNAME=m3-00 -e IOTLAB_GATEWAY_NO_INTEGRATION_TESTS --privileged iot-lab-gateway-test
```

aabadie: Fixed, the problem was on my side: I had to rebuild the containers. And this is a problem for me (non blocking, but will need to be fixed). Because this means each time I change code locally, I have to rebuild everything: is my local the gateway_code mapped in the container ? I thought it was.

rienafairefr: setup.cfg is not passed to the container, with `-v gateway_code:/home/iot-lab-gateway/gateway_code`  you mount only gateway_code , try `-v $CWD:/home/iot-lab-gateway`

aabadie: Ah yes !

rienafairefr: OK, I'm sure this would be more intuitive if we mounted the whole iot-lab-gateway folder instead of only gateway_code. Not sure about the implications

rienafairefr:
should we do code_check in CI ?

aabadie:
> should we do code_check in CI ?

Looking at the log output on travis it's already done and returns problems but [it always returns True](https://github.com/iot-lab/iot-lab-gateway/blob/master/tox.ini#L29)

aabadie:
I would use the real code_check return in a follow up task

aabadie:
Let's merge this, we'll see later for remaining things.

PR #13, opened 2017-10-11
Dockerize iot-lab-gateway env. 

rienafairefr:
The first purpose of dockerizing the gateway's REST application was to give an accessible development environment to actual and future developers.

Using the `docker-run` script, a user who built (docker-build) or has the gateway's docker image can either start a container with its own gateway_code directory mounted or not. This allows developers to make profit from a reloader option added to the REST application.


DOCKER.md
```diff
@@ -0,0 +1,77 @@
...
+
+The included Dockerfile includes all the necessary dependencies listed in `INSTALL.md`:
+
+* liboml2
+* openocd 0.9.0
+* iot-lab-ftdi-utils
+
+To build the image:
+
+    sudo .\docker-build
```
aabadie: Should it be `/` instead of `\` ? I see the same in other places

docker-run
```diff
@@ -0,0 +1,80 @@
...
+# The fact that you are presently reading this means that you have had
+# knowledge of the CeCILL license and that you accept its terms.
+
+# Use Docker-run
+# to run the docker container built by Docker-build.sh
+
+import argparse
+from collections import namedtuple
+
+import os
```
aabadie: This should be the first import. Apart from that, this file is full of pep8 and is not using common practices regarding a python script directly run from cli. You should use:
```python
if __name__ == '__main__':
   ...
```
at the end of the file and maybe define some utility functions.

gateway_code/config.py
```diff
@@ -127,3 +127,17 @@ def read_config(key, default=IOError):
...
+        :param key: Write configuration for 'key'
+        :param value: Value to save
+        :raises IOError: when 'key' can't be written """
+
+    entry = os.path.join(GATEWAY_CONFIG_PATH, key)
+
+    if value is not None:
+        with open(entry, 'w') as _conf:
+            _conf.write(value)
+
```
aabadie: There's a missing newline at the end of file (or more probably, trailing whitespaces on the last line)

gateway_code/gateway_logging.py
```diff
@@ -28,16 +28,19 @@
 from logging.handlers import RotatingFileHandler
 
 # set default logger level to DEBUG to log everything
+import sys
```
aabadie, line 31: This sys module is not used in this module, it should be removed

gateway_code/rest_server.py
```diff
@@ -55,6 +55,7 @@ def __init__(self, gateway_manager):
         self.board_config = board_config.BoardConfig()
         self._app_routing()
 
+
```
aabadie: This extra line is not pep8 compliant

gateway_code/tests/rest_server_test.py
```diff
@@ -205,13 +205,6 @@ def test_flash_function(self):
         ret = self.server.post('/open/flash', upload_files=[])
         self.assertEquals(1, ret.json['ret'])
 
-    def test_flash_idle(self):
```
aabadie: This looks like a rebase issue (are maybe no rebase against lastest master) ?

setup.py
```diff
@@ -71,7 +71,6 @@ def get_version(package):
 
 
 SCRIPTS = glob('bin/scripts/*')
-SCRIPTS += ['control_node_serial/control_node_serial_interface']
```
aabadie: What is the reason for this ? (I can't remember)

tests/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+# Update
+RUN apt-get update
+RUN apt-get install -y python-pip
+
+# Install app dependencies
+RUN pip install --upgrade pip
+
+RUN pip install tox
+
+CMD [ "tox","--develop" ]
```
aabadie: Missing newline (not sure this is very important though but GitHub reports this)

tests_utils/check_license.sh
```diff
@@ -14,6 +14,8 @@ files_list=$(echo "${files_list}" | grep -v \
     -e '.md$' \
     -e '.elf' \
     -e '.hex' \
+    -e '.travis.yml' \
+    -e 'Dockerfil' \
```
aabadie: s/Dockerfil/Dockerfile/

docker-build
```diff
@@ -0,0 +1,24 @@
...
+# with a limited warranty  and the software's author,  the holder of the
+# economic rights,  and the successive licensors  have only  limited
+# liability.
+#
+# The fact that you are presently reading this means that you have had
+# knowledge of the CeCILL license and that you accept its terms.
+
+import os
+
+os.system('docker build -t fitiotlab/iot-lab-gateway .')
```
aabadie, line 24: Do we really need another command line wrapper just for this ? It's be enough to mention it in the README.

aabadie (changes requested):
Before merging, I wanted to have a look at the changes here. See the comment below.

I think the introduced command lines (docker-gateway-rest-server and docker-run) could be refactored to be more compliant with python style command line tools.

Is this branch rebased on the very latest version of master ?

The possibility to log the gateway messages in stdout has been introduced, it would be nice, if possible, to update the `gateway_logging_test.py` regarding this change.

DOCKER.md
```diff
@@ -49,3 +49,25 @@ You can have the gateway run in the background by using a `-d` or `--daemon` arg
...
+
+Docker for Mac does not support adding a host device to a container (e.g. `docker run --device`). [source](https://docs.docker.com/docker-for-mac/faqs/#can-i-pass-through-a-usb-device-to-a-container)  
+The use of [Docker Toolbox](https://docs.docker.com/toolbox/overview/) is recommended.
+
+Start Docker Toolbox with the Docker Quickstart Terminal, then do this additionnal steps on the boot2docker VM used by Docker Toolbox:
+1. Install udev rules.
+
+        $ docker-machine ssh
+        docker@default:~$ git clone https://www.github.com/iot-lab/iot-lab-gateway.git && cd iot-lab-gateway
+        docker@default:~$ sudo cp bin/rules.d/* /etc/udev/rules.d/.
```
rienafairefr, line 85: @vandaele  Can we simply use python setup.py udev_rules_install here ? Hmm, maybe the boot2docker has no python

docker-build
rienafairefr: Well, in that case it's important the the built image is named fitiotlab/iot-lab-gateway, so that it can be used by docker-run, 

```./docker-build``` instead of     ```docker build -t fitiotlab/iot-lab-gateway .```, it's shorter and we are lazy.

Though it could be cool to know the docker build command, because ```docker build -t fitiotlab/iot-lab-gateway tests``` builds a slightly different image in which tests can be run, for example:

    docker build -t fitiotlab/iot-lab-gateway tests
    ./docker-run tox

DOCKER.md
vandaele: That's it

    docker@default:~$ python
    -sh: python: not found

docker-run
rienafairefr: OK mostly taken into account in commit 15e59781a00024da7a5d2878be3f92e6eb2214b7

gateway_code/config.py
rienafairefr: fixed in 15e59781a00024da7a5d2878be3f92e6eb2214b7

gateway_code/gateway_logging.py
rienafairefr: it seems to be used, sys.stdout is used for the stdout logging

gateway_code/rest_server.py
rienafairefr: I concur, fixed

setup.py
rienafairefr: control_node_serial/control_node_serial_interface is a binary. In order to be able to use tox we need to be able to make a sdist, in order to have this sdist containing that binary we need it declared in the package . Previously it was declared as a script. But it's not a script, it's a binary, so I've moved it to package_data (see below)

rienafairefr:
I've just this branch rebased on master

gateway_code/tests/rest_server_test.py
rienafairefr: OK, it was a rebase issue, I've put it back

PR #18, opened 2017-11-14
Comprehensive syntax code check

rienafairefr:
When starting tox, this will 
* start nosetests and fail the build if this fails
* examine the source code with pylint and flake8 and fail on any syntax problem

Failing a build because of a syntax problem might be pedantic, discussions welcome

### DIFF
--- a/roomba/roomba.py
+++ b/roomba/roomba.py
@@ -725,7 +725,7 @@ class Roomba500(object):
                 except KeyError:
                     pass
 
-            except:  # noqa: E722
+            except:
                 pass
 
             sens = self.get_sensor()
@@ -769,7 +769,7 @@ class Roomba500(object):
             try:
                 if self.qsens.qsize > 0:
                     self.qsens.get_nowait()
-            except:  # noqa: E722
+            except:
                 pass
             # Put the value of all sensors and position
             # of the Roomba with Queue
--- a/roomba/roomba.py
+++ b/roomba/roomba.py
@@ -725,7 +725,7 @@ class Roomba500(object):
                 except KeyError:
                     pass
 
-            except:
+            except:  # noqa: E722
                 pass
 
             sens = self.get_sensor()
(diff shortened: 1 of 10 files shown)


tox.ini
```diff
@@ -27,15 +27,14 @@ deps = -rtests_utils/test-requirements.txt
 commands =
     python setup.py nosetests {posargs}
```
aabadie: This toxenv looks useless to me since code_check and test envs already do the same things.

tox.ini
```diff
@@ -27,15 +27,14 @@ deps = -rtests_utils/test-requirements.txt
...
     bash -c "python setup.py lint | tee pylint.out || true"
-    bash -c "python setup.py pep8 | tee pep8.out || true"
-
+    bash -c "flake8 | tee flake8.out || true"
```
aabadie: I would keep the pep8 and add the flake8 check

roomba/roomba.py
```diff
@@ -769,7 +769,7 @@ def interface_serial(self):
             try:
                 if self.qsens.qsize > 0:
                     self.qsens.get_nowait()
-            except:
+            except:  # noqa: E722
```
aabadie: I think this can be excluded globally from the setup.cfg file. What is the humanly readable error type associated to it?

aabadie (changes requested):
Good initiative ! Just a few comments below.

tox.ini
rienafairefr: I agree, it was gaetan's way. I'll remove it if we are two to think it's useless :-)

roomba/roomba.py
rienafairefr: OK. I'll add 

    # E722 do not use bare except
    ignore = E722

 in flake8 config

tox.ini
rienafairefr: done, it should pass in travis

aabadie (approved):

aabadie:
All good here. Let's merge this. Thanks @rienafairefr !

rienafairefr:
"Your code has been rated at 10.00/10", et pas le droit de baisser en qualité :-)

aabadie:
> "Your code has been rated at 10.00/10"

with the help of `#pylint:disable` tricks ;)

PR #19, opened 2017-11-16
dependency injection

rienafairefr:
During development of dockerization, I was frustrated with the way config is passed to the gateway code, through a "/var/local/config"-like folder, which is hard to dockerize, hard to mock, hard all-around

This implements dependency injection for the board_type, control_node_type, hostname, robot, everything. The usual entrypoint for normal gateways (with the /var/local/config mounted from srvnfs etc) still works exactly the same.


aabadie:
Out-of-date and we won't use it. Closing.

PR #20, opened 2017-11-16
travis: simplify local and travis testing via docker image

aabadie:
This PR is an attempt to simply the use of docker image to run the test suite locally.

The idea is basically to share all the source code located on the host with the container. The current status in master is to map only the `gateway_code` directory which leads to very strange artefacts.

I also added a Makefile target that wraps the call to docker for testing.


aabadie:
I'm also wondering if we could move the Docker image build process somewhere and only pull the release image. I can see some limitations if people want new tools in the image (we'll then have to update the Dockerfile and build a new image) but it'll be very faster.

rienafairefr:
I've done some work in this direction as well, docker-di-cfg branch on my fork. It's building on #19 to reduce the test command to `docker run --privileged iot-lab-gateway-tests` :-) and it's green https://travis-ci.org/rienafairefr/iot-lab-gateway/builds/303044802 
If we merge #19  and docker-di-cfg we can have a very simple test procedure, check it out.

normal docker cache should take care of the slowness of the build, as long as we're pushing the image when it's building successfully. RIght now, we're not pushing fitiotlab/iot-lab-gateway to the docker hub when it's successfully built because we don't store credentials in travis. I'll look at the docker steps and why the cache is not reused properly when re-building

aabadie:
> I've done some work in this direction as well, docker-di-cfg branch on my fork

I was not aware of that, sorry

> If we merge #19 and docker-di-cfg we can have a very simple test procedure, check it out.

Sure I will. I had a brief look at #19 (only) and it obviously goes into the right direction.

> normal docker cache should take care of the slowness of the build, as long as we're pushing the image when it's building successfully

I was unsure if we really want to rebuild the docker image each but using caching technique will indeed help. This is also more flexible for users who want to contribute new open nodes (they can update the base image to their needs).

More information on the actual status of this PR:
I was not really convinced by the way the test command was run in the container: it was forcing me to rebuild the image before each run and it can be pretty confusing. Here, I tried to base the tests on the tests image but in the test image, the `WORKDIR` is different and is shared with the current directory on the host. Using a different entrypoint for mapping the host user ID in the container, it's now possible to either run tox within the container (if you don't have all dependencies installed locally) or just from the host.
With this and #19 and #21, we will be able to run integration tests on a board plugged on a laptop (I still don't remember how to do that).

aabadie:
Rebased on top of #43 to fix the CI

Makefile
```diff
@@ -0,0 +1,30 @@
...
+	sudo udevadm control --reload-rules
+
+test:
+	docker run -t --rm \
+		-v $(PWD):/shared \
+		-e LOCAL_USER_ID=`id -u $(USER)` \
+		iot-lab-gateway-tests tox
+
+integration-test:
+	echo $(BOARD) > $(PWD)/tests_utils/cfg_dir/board_type
```
rienafairefr: This is problematic because you're modifying the code source when testing, I would create the cfg_dir in /tmp maybe, copied from tests_utils/cfg_dir and modify it there

Makefile
```diff
@@ -0,0 +1,30 @@
...
+	docker run -t --rm \
+		-v $(PWD):/shared \
+		-v /dev/iotlab:/dev/iotlab \
```
rienafairefr: To avoid the --privileged, it would be best to do 

    --device /dev/ttynode    (the actual node, not the symlink)
    -v /dev/bus/usb/busnb/devicenb:[...]

the makefile could determine which /dev/tty and which /dev/bus/usb to share using `readlink`

tests/Dockerfile
```diff
@@ -27,4 +27,23 @@ RUN apt-get update && \
...
+    && curl -o /usr/local/bin/gosu.asc -SL "https://github.com/tianon/gosu/releases/download/1.4/gosu-$(dpkg --print-architecture).asc" \
+    && gpg --verify /usr/local/bin/gosu.asc \
+    && rm /usr/local/bin/gosu.asc \
+    && chmod +x /usr/local/bin/gosu
+
+COPY entrypoint.sh /usr/local/bin/entrypoint.sh
+
+WORKDIR /shared
+# Put control_node_serial_interface binary in the system path
+ENV PATH="/shared/control_node_serial:${PATH}"
```
rienafairefr: is that really needed ?

Makefile
aabadie: There's a tradeoff here: either we use the generic path and there's no need to pass the device path but you need the `--privileged`, either we use your proposition but then there's more logic to write in the Makefile and the user need to know the tty name on the host (in any case). I'd be in favour of the former for simplicity.

Makefile
aabadie: I know, but I find this very when testing.

> I would create the cfg_dir in /tmp maybe, copied from tests_utils/cfg_dir and modify it there

Good idea, I'll do that.

tests/Dockerfile
aabadie: Good catch, this is a leftover from previous debugging tests I made. It's not required.

aabadie:
Here is an update about the status of this PR:
* using a non root user inside the container works
* a helper Makefile is introduced with the following targets (use with `make <target>`)
  *  `setup-udev-rules`: install udev rules on the developper host
  *  `build-docker-image`: build base iot-lab-gateway docker image
  * `build-docker-image-test`: build test docker image and depends on `build-docker-image`. So using `make build-docker-image-test` will build both images in a row
  * `test`: run static and unit tests in the container
  * `integration-test`: run integration test on a board plugged on the host. Udev rules should be installed first. Use `BOARD=<open-node>` to specify a custom open node (default is M3). Example:
    <pre>
        make BOARD=arduino_zero integration-test
    </pre>

The idea is to keep this as simple as possible for users. By the way, I don't know `setup-udev-rules` will work on a Mac.

If this solution is accepted, we'll have to update the documentation.

rienafairefr:
This looks good. Even though we still require `--privileged` and having the udev rules installed on the host, at least the docker container doesn't pollute the current folder with root-owned files.

Makefile
```diff
@@ -0,0 +1,33 @@
+BOARD ?= m3
+
+.PHONY: build-docker-image build-docker-image-test setup-ude-rules \
```
aabadie: just notice that there's a typo on `setup-ude-rules` (missing `v` in ude**v**): make autocompletion spotted this.

rienafairefr: I saw that, didn't know where the typo was :-)

aabadie: it's now fixed

aabadie:
@rienafairefr, do you plan to work on the documentation ? Do you think it should be part of this PR or done in a follow-up one ?

PR #21, opened 2017-11-16
test: replace nosetest with pytest

aabadie:
This PR switch the testing tool from nosetests (not maintained anymore) to pytest.

It works when run locally but it needs further testing (not done yet) with integration tests.


aabadie:
This PR is based on #20

rienafairefr:
Great !

rienafairefr:
Are you still working on the host user right thing ? I feel like this pytest PR could be decoupled from the host user/gosu ? I can do it if you need

aabadie:
> I feel like this pytest PR could be decoupled from the host user/gosu ? I can do it if you need

Yes that would be great. Regarding the host user (see #20), I have problems that might be worth discussing (next week). The branch there is not up-to-date

rienafairefr:
I can push to your fork to update that PR, pytest seems to work in local. like you want.

aabadie:
> I can push to your fork to update that PR

Yes go ahead, thanks !

rienafairefr:
done, just need to verify that codecov gets the uploaded results with the pytest-generated coverage xml

aabadie:
Codecov seems to work

aabadie:
pytest 3.3 output is not as nice as locally

aabadie:
@rienafairefr, I spent some time making the switch to pytest fully work with integration tests (on real hardware). The last commit does this. 
Regarding coverage, there's a problem that is already in master. No report have been pushed for a moment. Will open a PR for this.

If code changes are ok for you, this PR can be merged.

tox.ini
```diff
@@ -32,37 +38,37 @@ whitelist_externals = /bin/bash
...
 whitelist_externals =
     /bin/bash
-    /bin/hostname
-# Use only sitepackages on gateways, no 'deps'
-# Only install 'webtest' as I'm lazy to package it for the moment
-deps = WebTest
-# Requires tox >= 1.8
-sitepackages = True
+    /usr/bin/make
+deps = -rtests_utils/test-requirements.txt
```
rienafairefr, line 48: the WebTest dependency is not used for anything in the end ?

aabadie: It's already installed by the `test-requirements.txt`. The thing is that before it was using the one that is packaged by yocto. I think it was to be as close as possible as a real environment.
When using the test-requirements.txt file, testing dependencies are installed in the tox env. It's simpler this way I think but maybe there are other things I missing.

rienafairefr:
Looks good, let's merge :+1:

aabadie:
I'm running a remote integration test on the m3 in devgrenoble. It works well.

PR #23, opened 2017-11-21
fix control node flash No

rienafairefr:


aabadie:
Looks good and the tests pass. Merging

PR #25, opened 2017-11-21
gateway: make openocd program path configurable for open nodes

aabadie:
The idea behind this PR is to give the possibility to configure the path of the openocd used to flash an open node.
New Yocto images for IoT-Lab have openocd 0.10 installed in `/opt/openocd-0.10.0/bin/openocd` so the idea is to have an optional option that can be defined in the open node class:
```python
OPENOCD_PATH = '/opt/openocd-0.10.0/bin/openocd'
```

The default is simply "openocd" as before.

I also applied the optional strategy to `OPENOCD_OPTS` attribute since it's only used for samr21 open nodes.

The integration tests still pass locally on an M3.

### DIFF
--- a/gateway_code/utils/openocd.py
+++ b/gateway_code/utils/openocd.py
@@ -30,18 +30,21 @@ import subprocess
 import atexit
 
 import logging
+from collections import namedtuple
 
 from gateway_code import common
 from . import subprocess_timeout
 
 LOGGER = logging.getLogger('gateway_code')
 
+OpenOCDArgs = namedtuple("OpenOCDArgs", ['path', 'config_file', 'opts'])
+
 
 class OpenOCD(object):
     """ Debugger class, implemented as a global variable storage """
     DEVNULL = open(os.devnull, 'w')
 
-    OPENOCD = ('openocd --debug=0'
+    OPENOCD = ('{openocd_path} --debug=0'
                ' {config}'
                ' -c "init"'
                ' -c "targets"'
@@ -60,8 +63,10 @@ class OpenOCD(object):
     DEBUG = ' -c "reset halt"'
     TIMEOUT = 100
 
-    def __init__(self, config_file, opts=(), verb=False, timeout=TIMEOUT):
-        self.config = self._config(config_file, opts)
+    def __init__(self, openocd_args,
+                 verb=False, timeout=TIMEOUT):
+        self.openocd_path = openocd_args.path
+        self.config = self._config(openocd_args.config_file, openocd_args.opts)
         self.timeout = timeout
 
         self.out = None if verb else self.DEVNULL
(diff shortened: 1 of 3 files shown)


rienafairefr:
"gateway_code/utils/openocd.py:63: [R0913(too-many-arguments), OpenOCD.__init__] Too many arguments (6/5)"

aabadie:
thanks, next time I'll also run the code_check locally ;)

aabadie:
> "gateway_code/utils/openocd.py:63: [R0913(too-many-arguments), OpenOCD.init] Too many arguments (6/5)"

@rienafairefr are you ok to use `# pylint:disable=too-many-arguments` here ?

rienafairefr:
Personnally, I'd refactor it. I'll send a PR your way in a sec

aabadie:
> I'll send a PR your way in a sec

What change do you propose ? (before PRing, worth discussing alternatives)

rienafairefr:
Instead of having multiple OPENOCD_, OPENOCD_, OPENOCD_  attributes in the node_ classes, having just one of type OpenOCDArgs that contain path, config_file, opts ?

aabadie:
> having just one of type OpenOCDArgs that contain path, config_file, opts ?

I also thought of that but this would require to change all existing open nodes. I wanted to keep the actual change minimal.
But it's tempting to have a namedtuple for that. I can update

rienafairefr:
woups, I meant to push to a fork, not your pr directly

aabadie:
> I meant to push to a fork, not your pr directly

let me know next time you do that. So we are not working both on the same thing...

> But it's tempting to have a namedtuple for that. I can update

maybe you missed the comment ?

rienafairefr:
I read the first part but not the part where you said you were going to update it, sorry :-1:

aabadie:
np, push you change. just read briefly and they were ok (happy to see we were doing the same). I was looking at a way to pass this to the open nodes.

rienafairefr:
okay, incoming

gateway_code/utils/openocd.py
```diff
@@ -30,18 +30,21 @@
...
 
 import logging
+from collections import namedtuple
 
 from gateway_code import common
 from . import subprocess_timeout
 
 LOGGER = logging.getLogger('gateway_code')
 
+Openocdargs = namedtuple("OpenOCDArgs", ['path', 'config_file', 'opts'])
```
aabadie: but rename it OpenOCDArgs (not nice but I think better)

rienafairefr: I had it with that case but code_check didnt like it for some reason

rienafairefr:
One vote to merge from me :-)

aabadie:
> One vote to merge from me :-)

not yet, I would like to check locally integration tests

gateway_code/utils/openocd.py
aabadie: I didn't get any error lint/pep8 error related to the use of OpenOCDArgs. I pushed an updated version. Let's see if travis return the same results.

aabadie:
@rienafairefr, Travis CI is green. (Rebase &) Merge at will !

rienafairefr:
ok closed #25

PR #26, opened 2017-11-22
IOTLABGATEWAY_CFG_DIR fix

rienafairefr:
in order to test with tox -e local with non m3 boards, it's necessary to keep the IOTLABGATEWAY_CFG_DIR to pass it to the tox env instead of overriding with tests_utils/cfg_dir 
This is a temporary hack. This is not necessary once the dependency injection PR is merged

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -52,7 +52,9 @@ deps = -rtests_utils/test-requirements.txt
 whitelist_externals =
     /bin/bash
 commands =
-    bash -c "export IOTLAB_GATEWAY_CFG_DIR={posargs:tests_utils/cfg_dir/}; \
+    bash -c "if [ -z '$IOTLAB_GATEWAY_CFG_DIR' ]; then \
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -52,7 +52,9 @@ deps = -rtests_utils/test-requirements.txt
 whitelist_externals =
     /bin/bash
 commands =
-    bash -c "export IOTLAB_GATEWAY_CFG_DIR={posargs:tests_utils/cfg_dir/}; \
+    bash -c "if [ -z '$IOTLAB_GATEWAY_CFG_DIR'  ]; then \
```
aabadie: still nitpicking: there's an extra space before `]`

rienafairefr: good catch

aabadie:
I think it's useful. And that also explains why I had to rebuild the docker image each time I changed the `test_utils/cfg_dir`

aabadie:
I tested this PR with success. Merging !

PR #27, opened 2017-11-23
Add support for ST LRWAN1 open node

aabadie:
This node is based on the ST discovery board [b-l072z-lrwan1](http://www.st.com/en/evaluation-tools/b-l072z-lrwan1.html).

Local integration tests are nearly working using the docker image.


rienafairefr:
Just tested on newly arrived STM32 LoRa, 

"Fatal Error in tests, stop further tests: "echo failed. Can't communicate with open node"

Did you manage to run some tox -e local on your side ?

aabadie:
> Did you manage to run some tox -e local on your side ?

Not fully, my only remaining issue is with the simple experiment test and looks more like a timing issue. I'm still looking into this.

aabadie:
After a very very hard time, I finally got this to pass the integration tests. Only "Simple experiment test" was failing and it was not because the port was wrong but because there's an issue with the management of the serial redirection process: when resetting the node, the serial redirection cannot be stopped properly => it was always returning with a non 0 code (only sigkill signal could stop it). I tested a lot of things (mainly increasing timings) and none of them worked.
So I decided to modify this particular integration test, see 0d883114853c04be467c44a28b760e6f520f1ef2. The idea is simply to ignore the error code of the serial redirection manager when it's stopped.

Appart from that, this open node works well with the 0.10 version of openocd.

aabadie:
@rienafairefr, do you think you'll have some time for testing this one ? I just realized the udev rules was not working and I pushed a fix (directly squashed)

rienafairefr:
OK just tested, local integration tests are passing, yay !

aabadie:
This one can be merged at any time. It has been tested with success on devsaclay.

PR #29, opened 2017-11-27
opennodes: use consistent idle firmware naming scheme

aabadie:
A small cleanup in to use a consistent naming scheme for open nodes firmwares. Some were having opennode_idle.elf/opennode_autotest.elf (samr21, arduino-zero) and some were having idle_opennode.elf/openode_autotest.elf (m3, leornado).

I think the first version having opennode_{idle,autotest}.elf is better as it groups firmwares related to a given opennode.

Normally all occurences have been renamed. If we merge this PR, we'll also have to think to update the [documentation on the website](https://www.iot-lab.info/hardware/custom/#contribute) (the inline image has the old version and should be updated).


rienafairefr:
Looks good to me

aabadie:
@rienafairefr, maybe we can merge this one ?

rienafairefr:
Yes, let's merge !

PR #30, opened 2017-11-28
Standardize nodes tty :  /dev/ttyON, /dev/ttyCN

rienafairefr:
I think this modification by itself would be alright, independently from PR #19 where it originated.

This puts the /dev/ttyON* /dev/ttyCN* inside a folder, /dev/iotlab, this makes it simpler to share devices when running the gateway code inside the docker container. Not sure it makes sense

-v /dev/iotlab:/dev/iotlab

instead of 

-v /dev/ttyON_FIREFLY:/dev/ttyON_FIREFLY


aabadie:
well, in both cases this can only work if the user has already installed on his host the udev rules of its open node

rienafairefr:
True. For a gateway with a single open node anyway, why not /dev/ttyON for any plugged node, and then

 -v /dev/ttyUSB0:/dev/ttyON

Thinking out loud

aabadie:
> -v /dev/ttyUSB0:/dev/ttyON

This is what I'm doing as well but it needs to documented somewhere.

vandaele:
Installing dev rules on the developper host should not be mandatory. I you are used to find your samr21 at /dev/ttyUSB0, and have already scripts using it for your usual work, you don't want that IoT-LAB change your habits and rename your samr21 device.

Otherwise, I agree with the `/dev/ttyON` idea.

rienafairefr:
I think I've covered all bases, I've updated the docker-run to handle the open node being /dev/ttyON , it greatly simplifies things (previously we had to do a dance with real path of the tty, manual udev, stuff like that), as a side-effect the docker-gateway-rest-server is removed. i've update the docs (DOCKER and DEVELOPER)

DOCKER.md
```diff
@@ -4,73 +4,66 @@ Docker Image of iot-lab-gateway
...
 
-    ./docker-run -v gateway_code
+    ./docker-run -v gateway_code /dev/ttyUSB0
 
 ## Examples
 
 You can use any type of open node:
 
-    ./docker-run -b samr21 -h samr21-test1
+    ./docker-run -b samr21 -h samr21-test1 /dev/ttyUSB0
```
aabadie: Here it should be /dev/ttyACM0 (for samr21)

bin/rules.d/a8-m3.rules
```diff
@@ -1,4 +1,4 @@
 # Open A8 M3 node
-SUBSYSTEM=="tty", SUBSYSTEMS=="usb", ENV{ID_SERIAL}=="IoT-LAB_A8-M3",         ENV{ID_USB_INTERFACE_NUM}=="01",  SYMLINK+="ttyA8_M3"
+SUBSYSTEM=="tty", SUBSYSTEMS=="usb", ENV{ID_SERIAL}=="IoT-LAB_A8-M3",         ENV{ID_USB_INTERFACE_NUM}=="01",  SYMLINK+="ttyON"
```
aabadie, line 2: I think we shouldn't change that for M3 nodes that are behind A8: this is because iotlab users may have written their own scripts using /dev/ttyA8_M3. If we change that, their scripts will be broken.

For other nodes, this is fine since the tty is only used by the gateway code.

docker-run
```diff
@@ -57,38 +46,51 @@ def main(args):
...
+    if input_str != '':
+        return ('-v {}:/home/iot-lab-gateway/gateway_code'
+                .format(os.path.abspath(input_str)))
+    else:
+        return ''
+
+
+def tty_type(prefix):
+    def func(input_str):
+        print('input_str %s'%input_str)
```
aabadie: Remaining print here.

aabadie (changes requested):
One minor things and a comment about the /dev/ttyA8_M3 that may break compatibility for users of A8 nodes.

DOCKER.md
rienafairefr: OK, fixed the doc

bin/rules.d/a8-m3.rules
rienafairefr: You're right, hadn't thought about that.

docker-run
rienafairefr: removed :+1:

rienafairefr:
Thanks @aabadie for the review

PR #31, opened 2017-11-29
opennode/arduino-zero: update node type string

aabadie:
As discussed in https://github.com/iot-lab/iot-lab-dev/pull/28, I'm renaming the node type to `arduino-zero`, it will be nicer for users to use `arduino-zero-1` instead of `arduino_zero-1`

### DIFF
--- a/gateway_code/open_nodes/__init__.py
+++ b/gateway_code/open_nodes/__init__.py
@@ -37,7 +37,8 @@ def node_class(board_type):
     :raises ValueError: if board class can't be found """
     try:
         module_path = '%s.%s' % (
-            __name__, OPEN_NODES_MODULE.format(type=board_type))
+            __name__, OPEN_NODES_MODULE.format(
+                type=board_type.replace('-', '_')))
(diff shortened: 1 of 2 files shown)


aabadie:
I'll also update the tests. Please wait before merging.

rienafairefr:
You need to add .replace('-', '') to _node_title to fix

aabadie:
Testing locally a fix, will push if it works.

aabadie:
it works, what do you think @rienafairefr ?
basically, the only problem is that the open node type, as defined in the board class, should always use `_` but in the resources file, it can contain a `-`.

rienafairefr:
We will probably soon move to where the module name and node names/types don't have to be correlated so I think it's not a big problem, your fix is fine

aabadie:
Let's merge this and it will be in the nightly built image. Then I can move forward with https://github.com/iot-lab/iot-lab-dev/pull/28 and use the name with `-`

rienafairefr:
Okay

PR #32, opened 2017-11-29
firefly bug with tty case

rienafairefr:
The firefly support was not following the ttyON_NODENAME convention


aabadie:
Missed this one and it looks ok, let's merge it.

PR #33, opened 2017-11-29
.idea gitignore

rienafairefr:
Sorry the .idea/ litters the gitignore in multiple repos, but it's a directory that is used by pycharm


aabadie:
Maybe use a more explicit commit message ?

rienafairefr:
done :+1:

aabadie:
Thanks, if CI is green, we can merge then !

PR #34, opened 2017-11-29
Simplify Travis

rienafairefr:
Simplify the docker build/run in travis ci.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,11 +20,4 @@ before_cache:
   - docker save -o docker/previous_image.tar.gz iot-lab-gateway
 
 script:
-  - >
-    docker run -v gateway_code:/home/iot-lab-gateway/gateway_code
-    -v $CWD/tests_utils/cfg_dir:/var/local/config
-    -e BOARD_TYPE=m3
-    -e CONTROL_NODE_TYPE=no
-    -e HOSTNAME=m3-00 
-    -e IOTLAB_GATEWAY_NO_INTEGRATION_TESTS
-    --privileged iot-lab-gateway-tests
+  - docker run -t iot-lab-gateway-tests tox
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -27,4 +27,4 @@ RUN apt-get update && \
 RUN pip install --upgrade pip && \
     pip install tox
 
-CMD [ "tox","--develop" ]
+CMD [ "tox" ]


aabadie:
Can you explain a bit more what's the benefit of this PR ?

Since only the basic tests can be run on travis, why not just use tox outside the container ?

rienafairefr:
The previous .travis.yml was written with lot of added cruft and options added to docker run 'just in case', which are, now we know, not needed (all the environnement variables for example) and it's ugly

tox outside the container means installing all the dependencies (like oml) in travis. Which are already installed inside the container, that would mean duplication

.travis.yml
```diff
@@ -20,11 +20,4 @@ before_cache:
...
 script:
-  - >
-    docker run -v gateway_code:/home/iot-lab-gateway/gateway_code
-    -v $CWD/tests_utils/cfg_dir:/var/local/config
-    -e BOARD_TYPE=m3
-    -e CONTROL_NODE_TYPE=no
-    -e HOSTNAME=m3-00 
-    -e IOTLAB_GATEWAY_NO_INTEGRATION_TESTS
-    --privileged iot-lab-gateway-tests
+  - docker run --privileged iot-lab-gateway-tests
```
rienafairefr: one small command to launch tests, what's not to love ;-)

aabadie: IIUC, the `--privileged` option is not required ?

rienafairefr: indeed, removed

.travis.yml
```diff
@@ -20,11 +20,4 @@ before_cache:
...
 script:
-  - >
-    docker run -v gateway_code:/home/iot-lab-gateway/gateway_code
-    -v $CWD/tests_utils/cfg_dir:/var/local/config
-    -e BOARD_TYPE=m3
-    -e CONTROL_NODE_TYPE=no
-    -e HOSTNAME=m3-00 
-    -e IOTLAB_GATEWAY_NO_INTEGRATION_TESTS
-    --privileged iot-lab-gateway-tests
+  - docker run iot-lab-gateway-tests
```
aabadie: Use `-t` for colored output. And I also prefer to have `tox` appended to this command and removed from the docker image.

rienafairefr: if I remove tox from the dockerfile, it might be conterintuitive to run `docker run iot-lab-gateway-tests` (without the tox) and having the gateway rest server starting IMO

rienafairefr: I'm not sure the -t will change anythign for travis but I'll add it ^^

aabadie: Apparently it does make a difference (that's why I asked, I already noticed that ;) )

tests/Dockerfile
```diff
@@ -27,4 +27,4 @@ RUN apt-get update && \
 RUN pip install --upgrade pip && \
     pip install tox
 
-CMD [ "tox","--develop" ]
+CMD [ "tox" ]
```
aabadie, line 30: Maybe simply use `[ "/bin/bash" ]` ?

aabadie:
Just a minor nitpick, please squash and we can merge this one.

aabadie:
> tox outside the container means installing all the dependencies (like oml) in travis. Which are already installed inside the container, that would mean duplication

Ok, I got it now. Thanks.

rienafairefr:
OK, consider it done

tests/Dockerfile
rienafairefr: I'll leave tox here, that way `docker run iot-lab-gateway-tests` will start it, instead of starting the gateway rest server (CMD inherited from the iot-lab-gateway docker image)

rienafairefr:
I've squashed, merging it now.

PR #35, opened 2017-11-30
micro:bit support

rienafairefr:
This is the basics of supportingMicro:bit board on FIT iot-lab

firmware is built as usual in ci-firmwares using RIOT

programming is done through pyOCD https://github.com/mbedmicro/pyOCD

tested locally without hiccups. Code needs squashing, I know :-)


aabadie:
Very nice, I have a few of these so I can test.

aabadie:
@rienafairefr I finally tested this PR locally, I ran the integration tests and all passed except the `Test simple experiment`. It seems to be the same issue I had for the STM32 LRWAN1 board: the serial redirection is not closed properly after a reset (a ret code != 0 is returned and the test expect 0). I fixed it by ignoring the return code which is not very nice but works.

See 5f12a9dc34f63df90a89cd59523d90c3d7666628 for details on the test update. The problem comes from [there](https://github.com/iot-lab/iot-lab-gateway/blob/master/gateway_code/utils/serial_redirection.py#L125) in the code. Previous signals (sigterm, sigint) failed to stop the serial redirection and the last attempt uses sigkill: the redirection is stopped but we get a non-zero return code.

Otherwise, I very much like this PR and the use of PyOCD (I never tried that tool). It will be awesome to have microbits on iotlab !

aabadie:
The integration tests obviously pass when applying 67337ce here.

aabadie:
Re-tested this PR again and now it works. Seems like you rebased on #20 and cherrypicked the commit about serial redirection from #27

rienafairefr:
Yes, this will not be necessary once #20 and #27 are merged.

aabadie:
@rienafairefr, just realized that the firmwares are 775, they should be 664 IIRC.

aabadie:
I've played a bit with pyocd and there's one drawback (not critical, for now): it doesn't work with python 3. I hope they'll do the move soon because python2 support will probably stop in a couple of years.

aabadie:
Ok, I wanted to merge this one but it seems to be based #20.

Should we merge #20 first or just take the board support from here ?

rienafairefr:
I rebased and remove the dependency on #20, we'll merge that one later.

rienafairefr:
I deployed CI nodes on devlille for this, if no voices against the merging this will be merged soon :-)

aabadie:
> if no voices against the merging this will be merged soon

One comment: it would be great to a test module for this new tool. That would be useful I think (only 41% of the diff is covered by tests).

rienafairefr:
I've added pyocd unit tests, upping the coverarge :-), merging !

PR #38, opened 2017-12-08
fix USER env not working ?

rienafairefr:
I don't know why this stopped working but using tox {env:KEY} instead of the $KEY seems to work.


fsaintma:
Tested on my computer and it works. I merge and good job. We will never know :)

PR #39, opened 2017-12-08
fix GCOV environement variable for tox -e control_node_serial

rienafairefr:


tox.ini
```diff
@@ -69,6 +69,7 @@ commands =
 skip_install = True
 sitepackages = True
 deps = gcovr
+setenv = GCOV=arm-poky-linux-gnueabi-gcov
```
aabadie: I don't think this will work when testing locally ? or maybe I'm missing something

rienafairefr: You're right.

PR #43, opened 2017-12-17
fix issues raised by pylint 1.8

aabadie:
This PR fixes new issues raised when using pylint 1.8.

See changelog of pylint: https://github.com/PyCQA/pylint/blob/master/ChangeLog


gateway_code/rest_server.py
```diff
@@ -310,7 +310,7 @@ def _cond_route(self, obj, func, path, *route_args, **route_kwargs):
 
     def route(self, path, method='GET', callback=None, *args, **kwargs):
         """Add a route but catch some exceptions."""
-        # pylint:disable=arguments-differ
+        # pylint:disable=arguments-differ, keyword-arg-before-vararg
```
rienafairefr, line 313: maybe we shouldnt disable the warning but address the *args, *kwargs used in all these _route methods in GatewayRest ? Most of them are not really needed IMO

PR #44, opened 2017-12-18
serial_redirection: add named option for setting a custom start delay

aabadie:
This PR is an alternative to https://github.com/schrein/iot-lab-gateway/pull/2 and https://github.com/schrein/iot-lab-gateway/pull/3

It doesn't change the behaviour of existing open nodes and is easy to adapt for Zigduino for example: just call `ret_val += self.serial_redirection.start(start_delay=2)` [here](https://github.com/iot-lab/iot-lab-gateway/blob/master/gateway_code/open_nodes/node_zigduino.py#L74).


aabadie:
Closing since it's apparently not the right fix for zigduino.

PR #45, opened 2018-01-12
Various fixes with node names/types containing '-'

aabadie:
This PR fixes the way the control nodes determines the node archi and num from the hostname (written in /var/local/config).

This PR also proposes a fix when the board type contains a '-' instead of '_'.

This problem happened with arduino-zero-x nodes available in Saclay site.


aabadie:
Merging this one, now

PR #47, opened 2018-02-06
utils: add support for edbg flash tool

aabadie:
Since this tool targets the SAMR21 and Arduino Zero open nodes, I updated their support accordingly.

This PR has not beed tested on hardware and requires https://github.com/iot-lab/iot-lab-yocto/pull/13

PR #48, opened 2018-02-06
utils: remove unused elf_path flash parameter

aabadie:
This is a minor fix in the CC2538 flash utility: the `elf_path` is not used by the flash command.


rienafairefr:
Good to merge directly :-)

aabadie:
I found that because I looked at this one when writing #47 :)

PR #53, opened 2018-04-10
update setuptools for installing pyocd

rienafairefr:
A recent upgrade of pyOCD or setuptools-scm or whatever broke something, we need to self-update pip and setuptools before installing pyOCD (the apt packaged ubuntu 14.04 version of setuptools is ancient). This failure was hidden by the docker cache mechanism in travis


aabadie (approved):
Looks good

Issue #55, opened 2018-04-16
Feature for Custom Standalone (M3) Control Node


rdagher:
Hello,

I have created a [gist](https://gist.github.com/rdagher/0adb14208a906f812eb7b7076b02b12a) and a [fork](https://github.com/rdagher/iot-lab-gateway/tree/stalone_cn_m3) for using the Docker image with **custom** control node, in this case a standalone M3 is used as a standalone control node just capable of sniffing the traffic.
Note that the fork is several commits behind master and some bugs were encountered in the control node native interface (wrapped by python classes). A test script is also available...

Best,
Roudy

rienafairefr:
Looks interesting.
- Can you push a pull request so we can see the differences ?
- In the code, looks like you're using the docker-run, but recently we've migrated to the Makefile (target `make run`), which is much easier than the docker-run script. Maybe rebase on our master and see that (we've updated the documentation)

rdagher:
Well while I was validating this, modifications were being made in the maser regarding the docker run. I will make a pull request so you can test this and then update with the new strategy.

PR #56, opened 2018-04-17
Standalone M3 Control Node

rdagher:
issue #55


rienafairefr:
I took the liberty to rebase on latest master, much clearer now :-)
While I look at this more, can you verify it still works for your setup ? (I had no major merge conflicts on the code though)

gateway_code/control_nodes/cn_iotlabstalone/__init__.py
```diff
@@ -0,0 +1,68 @@
+# -*- coding:utf-8 -*-
```
rienafairefr: stalone looks like a typo to me, I guess the directory should be named cn_standalonem3 ?

gateway_code/control_nodes/cn_iotlab/__init__.py
```diff
@@ -190,7 +190,7 @@ def _wait_control_node_ready():
         So wait 2 seconds to be safe.  """
         time.sleep(2)
 
-    @staticmethod
-    def status():
+    @logger_call("Control Node : status")
```
rienafairefr: Yeah, this looks fine, but this is not related to standalone M3 control node IMO. Put that un another PR, with other small fixes

gateway_code/control_nodes/cn_iotlabstalone/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+from gateway_code.utils.ftdi_check import ftdi_check
+from gateway_code.config import static_path
+
+from gateway_code.control_nodes.cn_iotlab import ControlNodeIotlab
+
+
+LOGGER = logging.getLogger('gateway_code')
+
+
+class ControlNodeIotlabstalone(ControlNodeIotlab):
```
rienafairefr: stalone everywhere, maybe it's not a typo in the end ?

gateway_code/control_nodes/cn_iotlab/__init__.py
rienafairefr: Well, it's actually breaking the linter, because the status method should be static (we don't use self in it). either ignore it (add appropriate pylint: ignore),

gateway_code/control_nodes/cn_iotlabstalone/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+    TYPE = 'iotlabstalone'
+    FEATURES = ['leds',
+                'radio']
+    OPENOCD_CFG_FILE = static_path('iot-lab-cn-stalone.cfg')
+
+    def __init__(self, node_id, default_profile):
```
rienafairefr: not needed, because your class is a subclass of ControlNodeIotlab

gateway_code/control_nodes/cn_iotlabstalone/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+        super(ControlNodeIotlabstalone, self).__init__(node_id, default_profile)
+
+    @logger_call("Control node : profile configuration")
+    def configure_profile(self, profile=None):
+        """ Configure the given profile on the control node """
+        LOGGER.info('Configure profile on Control Node')
+        self.profile = profile or self.default_profile
+        ret_val = 0
+
```
rienafairefr: plenty of weird lines with spaces at the end

tests_utils/test-cn-stalone-Docker.sh
```diff
@@ -0,0 +1,51 @@
+#! /bin/bash -x
```
rienafairefr: this should be rewritten to take into account the Makefile that has been added in the meantime

rienafairefr (changes requested):
Still some work to do :-) :+1:

rdagher:
Gr8. I made some updates, and it works on my setup. Check it out.

gateway_code/control_nodes/cn_iotlabstalone/__init__.py
aabadie: and why not simply `cn_iotlabm3` ?

gateway_code/utils/ftdi_check.py
```diff
@@ -21,17 +21,73 @@
...
+        found = (description == ftdi_parse_device_description(lines[i + 2]))
+        if found:
+            break
+    return found
+
+
+if __name__ == "__main__":
+    print ftdi_check('control', '2232')
+    print ftdi_check('control', '2232', 'ControlNode')
+    import doctest
```
aabadie: I think this is already called by nose (and soon pytest)

gateway_code/control_nodes/cn_iotlabstalone/__init__.py
rdagher: OK, Changed the name to `cn_iotlabm3 `.

gateway_code/utils/ftdi_check.py
rdagher: Ok. Removed the main.

rienafairefr (approved):

Makefile
```diff
@@ -46,9 +46,11 @@ run: setup-cfg-dir
 	docker run -it --rm \
 		-v $(PWD):/shared \
 		-v /dev/iotlab:/dev/iotlab \
+		-v /dev/ttyCN:/dev/ttyCN \
 		-v /tmp/cfg_dir:/var/local/config \
 		-p $(PORT):8080 \
 		-p $(SERIAL_PORT):20000 \
+		--name "gateway_test" \
```
rienafairefr: Yeah I think we shouldn't force docker container names, esp. if you re-do a docker run with the same name it will error. I removed 'gateway-test' name in 8c7f540475d736384891f793e30256533a0fd2b7 when deleting the docker-run

Makefile
```diff
@@ -46,9 +46,11 @@ run: setup-cfg-dir
 	docker run -it --rm \
 		-v $(PWD):/shared \
 		-v /dev/iotlab:/dev/iotlab \
+		-v /dev/ttyCN:/dev/ttyCN \
```
rienafairefr: @aabadie  mentioned it, but what happens when the host doesn't have /dev/ttyCN ?

tests_utils/test-cn-m3-Docker.sh
```diff
@@ -0,0 +1,48 @@
...
+# - Docker container named 'gateway_test' attached to a standalone M3 OpenNode and standalone M3 ControlNode
+# - The ControlNode board's FTDI eeprom has been set with the name ControlNode (using the IotLab ftdi-eeprom-config tool)
+
+# Experiment profile (sniffer) and firmware
+profile="../gateway_code/tests/profiles/radio_sniffer_3.json"
+firmware="../gateway_code/static/m3_autotest.elf"
+
+# Mocked user workspace and experiment folders (usually mounted by nfs)
+# here, they are created in the (alreadyà running container
+user="test"
```
rienafairefr: you should use or modify one of the makefile targets (I think `make run` will do what you need, just add a -v mapping a folder in /tmp to /iotlab/users/.../experiment)

rienafairefr (changes requested):
the docker script is still messy with the docker exec, try the make run I just modified

rdagher:
I don't understand why you want to avoid explicit container name, but I do understand the use of volumes to retrieve the experiment output. 
Updated the [test-cn-m3-Docker.sh](https://github.com/rdagher/iot-lab-gateway/blob/stalone_cn_m3/tests_utils/test-cn-m3-Docker.sh) to first run the server as suggested and dump sniffing result from the shared volume `/tmp/exp_dir/sniffer/`.

Makefile
rdagher: What happens is that the Docker daemon creates a folder `/dev/ttyCN` and mounts it !
Of course this is not clean. What about moving the symbolic link `ttyCN` to the `/dev/iotlab`folder? 
I can give it a shot (update rules and code) and test) if you agree...

rienafairefr:
The problem with named containers is that once a name is taken by a container you can't create another one with the same name, even if the conflict comes from a stopped container, so re-launching stuff can fail

Makefile
rienafairefr: Yeah I think the best would be to put ttyCN in /dev/iotlab. 
in udev SYMLINK+="ttyCN[separator]iotlab/ttyCN" would create both symbolic links, I think separator is a comma not sure

aabadie: I would prefer adding some Makefile logic that sets a variable with mapping if requested by the user:
```
make USE_CN=1 integration-test
```

In the makefile, test that USE_CN is equal to 1, default value is 0:
```mk
USE_CN ?=0

DOCKER_CN_MAPPING=
ifeq (1,$(USE_CN))
  DOCKER_CN_MAPPING = -v /dev/ttyCN:/dev/ttyCN
endif
```

And then in the integration-test target:
```mk
docker run -it --rm \
 		-v $(PWD):/shared \
 		-v /dev/iotlab:/dev/iotlab \
		$(DOCKER_CN_MAPPING) \
```

Not tested, but you have the idea.

rdagher: @aabadie I see the idea, the logic is actually in the variable `CONTROL_NODE_TYPE`. I am pushing a fix.

rienafairefr:
with latest additions, looks good to me

rienafairefr (approved):

rienafairefr:
@aabadie, I think I'm merging this today unless you have last-minute objections

aabadie:
FYI, this PR broke the integration tests. After digging into the issue on our CI, it appears that the control node on legacy gateways was not working as it should. In the end, the problem was that this PR modifies the firmware that is running on the control node and this breaks the control node on legacy IoT-LAB gateways.

So I pushed a commit that revert the original control node firmware. If you want to use a specific one for standalone M3 CN, please provide it separately. Ideally, you should also provide the source somewhere (not in this repo). Thanks !

rdagher:
@aabadie I had to recompile the firmware of the control node with the latest version from openlab. The reason I had to do that is that I had some issues with the control node interface, somme commands were missing I think, but i will test this again with the firmware you reverted. 
BTW, in production, this same firmware is used to flash the CN or is there any repo for CN images?

aabadie:
I'm not 100% sure but I think the source code of the application that runs on the control node is this [here](https://github.com/iot-lab/openlab/tree/master/appli/iotlab/control_node/control_node).
I don't think it's possible to use exact same application for the legacy CN and an M3 base CN, because the M3 base CN won't be able to control the alim of the open node.

Maybe you forgot to pass the `IOTLAB_CN` build flag when you rebuilt the new firmware version.
Now I think the best is to provide the firmware that run the M3 as CN in a separate file and adapt the `ControlNodeIotlabm3` python class accordingly.

PR #58, opened 2018-05-24
rewrite some udev rules

rienafairefr:
I think this will fix some problems with detection of devices, symlinks etc:
- A8 and A8-m3, in addition to the /dev/ttyON_A8, also create a symlink in /dev/iotlab, like other nodes. Tested on an A8 gateway
- rewrote microbit.rules, this is more of a cmsis-dap.rules, matching all cmsis-dap nodes (will include some future included nodes, I think). Tested on my machine
- I think ST_LRWAN1 had the same issue (symlink appearing when no st_lrwan1 is plugged), so I added idVendor and idProduct matching. Not yet tested on real hardware


aabadie:
I tested the changes on microbit and st-lrwan1 open nodes. All good there. For A8, I ran the integration test on devgrenoble and they seem to pass. The `/dev/iotlab/ttyON_A8` and `/dev/ttyON_A8` are also both there. I guess it's ok for the `A8_M3`.

Let's merge this one.

PR #59, opened 2018-06-08
integration_tests: skip no log error check for nodes with dirty serial

aabadie:
This PR provides a way to skip the no error check on serial for nodes that can sometimes have remaining data on the serial link.
This is typically the case for the ST-LRWAN1 node. Keeping this check is problematic since the CI fails otherwise. Here I suspect a more deep problem with the autotest firmware: the RIOT UART driver for stm32l0 could have a potential bug during initialization.

The proposed solution will just skip this check if the DIRTY_SERIAL attribute is set for a given open node. Tested on devsaclay and works.


fsaintma:
@aabadie the same like other PR should you resolve the conflict before merge this one.

PR #60, opened 2018-06-18
Fix issue with nodes flashed with edbg and debugging sessions

aabadie:
This PR provides a fix for SAMR21 and Arduino-Zero that doesn't work correctly when the user start a debugging session on them.
The problem started when the Edbg flasher tool was introduced: when the node is flashed with Edbg, any attempt to flash it via a GDB session (via openocd) after leave the node in a broken state.
But it works when the node was initially flashed with openocd.

So the fix here consists in tracking the current firmware that is running on the node at the beginning of the debugging session and reflash it with openocd before opening gdb.

The other solution is to drop edbg as flashing tool and use openocd instead but in this case, the flash procedure will become a lot slower.

Tested on CI node samr21 and arduino zero and all integration tests passed.

### DIFF
--- a/gateway_code/integration/integration_test.py
+++ b/gateway_code/integration/integration_test.py
@@ -242,11 +242,15 @@ class TestComplexExperimentRunning(ExperimentRunningMock):
         ret = self._flash()
         self.assertNotEquals(0, ret.json['ret'])
 
-        # No flash, Autotest fw should be still be running
+        # No flash, Autotest fw should be still running
         self._check_node_echo(echo=True)
 
         # Stop debugger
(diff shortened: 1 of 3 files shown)


fsaintma:
@aabadie I have just saw your PR and it seems to be good. Just resolve conflict in integration_tests and I will merge

aabadie:
> Just resolve conflict in integration_tests and I will merge

done, let's wait for travis, you can also test on one of the CI nodes (it will take a bit more time)

aabadie:
The 0% of diff hit is normal: this part of the code is not tested on Travis.

PR #63, opened 2018-06-25
tox: use sitepackages when running integration tests

aabadie:
This PR revert the use of dependencies from sitepackages when running the integration tests on a gateway. This PR now also makes use of WebTest from sitepackages instead of installing it in the tox environment.

**Warning:** this PR requires https://github.com/iot-lab/iot-lab-yocto/pull/19 to be merged and deployed on CI images first.

Since `pytest-pep8` is not packaged in the yocto image (do we want this ?), the `--pep8` option has been removed from the setup.cfg and is now passed explicitly to pytest when required.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,7 +3,6 @@ addopts =
     -v
     --doctest-glob="doc/*.rst"
     --doctest-modules
-    --pep8
     --cov=gateway_code
     --cov-report=xml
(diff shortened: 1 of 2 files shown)


aabadie:
https://github.com/iot-lab/iot-lab-yocto/pull/19 is merged and deployed on CI nodes

PR #67, opened 2018-07-05
open_nodes: use generic STLINK port for ST LRWAN1

aabadie:
This PR generalize the ST LRWAN open node associated udev rules to be st-link. This is because the same udev rules applies to boards with an st-link adapter (all nucleos and other discovery boards).

It's the same problem that was reported for boards having a DAP Link, like the microbit.


rienafairefr:
Looks fine :+1:

aabadie:
@rienafairefr, I reworked this PR and introduced an ST-Link base open node. Now the ST LRWAN1 node is based on it. I'll reuse it in #68 (already did that in a local branch).

aabadie:
Apparently pkgutil doesn't like relative imports...
```
ValueError: Attempted relative import in non-package
```
Already fixed and now running integration tests in devsaclay (smells good)

aabadie:
Integration tests passed. Let's wait for Travis and then I think we can merge.

PR #68, opened 2018-07-05
open-node: add support for ST B-L475E-IOT01A (ST-IOTNODE)

aabadie:
This PR adds another ST openode, the [B-L475E-IOT01A](https://www.st.com/en/evaluation-tools/b-l475e-iot01a.html). This boards embeds an STM32L4 CPU, low-power but not very constrained in terms of RAM (128kB) and ROM (1MB). It also has several types of radio: WiFi, BLE and SubGHz. And finally, it comes with a few sensors (IMU, pressure, humidity, temperature).

What else ? :)

This PR is also based on #67 since it's also using an st-link adapter.

It has already been tested with success on devsaclay (iotnode-1).


aabadie:
Integration tests have passed on devsaclay. Travis is green. I merge !

PR #70, opened 2018-08-10
Open nodes: add EDBG base class and use it with samr21 and arduino zero

aabadie:
This PR factorize the open nodes implmentation of the arduino-zero and samr21 boards.

It also does a small cleanup in the st-link base class.

Integration tests have passed for samr21 and arduino-zero.


aabadie:
CI is green, integration tests have passed. Let's merge.

PR #71, opened 2018-08-10
open_node: add support for nrf52dk

aabadie:
This PR add the [nRF52dk](https://www.nordicsemi.com/eng/Products/Bluetooth-low-energy/nRF52-DK) board as a new open node.
This board provides BLE radio.

~Integration tests only work with RPI3 based gateways but not with legacy IoT-LAB gateways (for an unknown reason, probably something related to the Linux kernel).
Wth RPI3 based gateways, it doesn't work out of the box though, one has to manually power cycle the node first. I'll open a PR soon that integrate this in the gateway code.~


aabadie:
After some fight with integration tests, I found what was the origin of the problem with JLink: the idle firmware flashed was always printing a welcome message during boot and this somehow breaks the communication with the serial redirection.

So I deactivated the RIOT welcome message in the ci-firmwares and all is working well now. The integration tests are passing on a legacy iotlab gateway. So I'll merge this one asap (as I need it).

PR #72, opened 2018-08-10
control_nodes: add RPI3 control node

aabadie:
For the moment it only provides the open-node-power feature of a plugged open node.

Unit tests have been added.

Work that remains:
- being able to run the tool with user www-data because for the moment I could only make it work as root on a gateway
- test it with integration test (requires update of resource file)
- test it with nrf52dk-2 in devsaclay and see if it fixes the serial link access problem for this open node

### DIFF
--- a/gateway_code/control_nodes/cn_rpi3/__init__.py
+++ b/gateway_code/control_nodes/cn_rpi3/__init__.py
@@ -1,0 +1,133 @@
+# -*- coding:utf-8 -*-
+
+# This file is a part of IoT-LAB gateway_code
+# Copyright (C) 2015 INRIA (Contact: admin@iot-lab.info)
+# Contributor(s) : see AUTHORS file
+#
+# This software is governed by the CeCILL license under French law
+# and abiding by the rules of distribution of free software.  You can  use,
+# modify and/ or redistribute the software under the terms of the CeCILL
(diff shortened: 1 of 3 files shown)


aabadie:
Further improvements:
- add LEDs control (to check if the gateway is idle or an experiment is running, etc)
- add a sniffer module
- add power measurement extensions, etc

aabadie:
Small update about the remaining work on this PR:

> being able to run the tool with user www-data because for the moment I could only make it work as root on a gateway

This is now possible with https://github.com/iot-lab/iot-lab-yocto/pull/21. It has been tested on nrf52dk-2 on devsaclay (which is running behind a rpi3 gateway).

> test it with integration test (requires update of resource file)

Integration tests runs fine on nrf52dk-2.

> test it with nrf52dk-2 in devsaclay and see if it fixes the serial link access problem for this open node

This PR didn't fully fixed the serial access problem. But this problem is now solved by using better idle/autotest firmware as described [here](https://github.com/iot-lab/iot-lab-gateway/pull/71#issuecomment-415369012).

In conclusion, this PR can be merged when https://github.com/iot-lab/iot-lab-yocto/pull/21 is in and deployed on a rpi3 gateway.

aabadie:
iot-lab/iot-lab-yocto#21 is merged, so let's go with this one.

PR #73, opened 2018-08-24
open_node: add support for nrf52840dk board

aabadie:
This board is very similar to nrf52dk (same MCUs) but it also provides 802.15.4 connectivity.

This PR also introduce a common base class for both nrf52dk and nrf52840dk nodes, since they share a lot things.


aabadie:
Integration tests passed on both nrf52dk and nrf52840dk nodes. CI is green here, let's merge this PR.

PR #74, opened 2018-08-24
static: update idle and autotest firmwares used by open nodes

aabadie:
I pushed a fix in the ci-firmware Makefile that prevents the autotest and idle firmware to display the RIOT welcome message. This was the cause of problems with nrf52 based boards but it's also not super nice to have this message displayed when one reset a node with no user firmware already flashed.

Autotest firmwares were not update for m3, fox and a8 because they are still using the openlab versions with more tested features.

The arduino-zero firmware were renamed with the same filename that is generated by the ci-firmwares make build.


aabadie:
I ran integration tests (only autotests and simple experiment) on all CI nodes (except zigduino), all have passed.

PR #75, opened 2018-08-28
tests: improve code coverage with unit tests

aabadie:
This PR provides several things:
- fix coverage miss in the `avrdude_test` module
- add new test function for the `elftarget` module
- add missing unit tests for the `cc2538` bsl module. For this one I suspect that debug_start and debug_stop cannot be implemented because `cc2538-bsl.py` is not able to launch a gdb server (like openocd). I think the `debug_start/debug_stop` methods should be removed from the cc2538.py module and also from the Firefly open node
- add a new test function for `serial_expect` (for the context manager feature)
- increase the coverage of the `node_connection` module
- add a test function for the debug command of `cli_openocd`
- add basic unit test to the control node no (`cn_no`)
- most important: introduce a new open node base class, `NodeOpenOCDBase` and use it with microbit, jlink, st_link, fox and m3 existing node.
- add unit test for `NodeOpenOCDBase` and `NodeOpenEdbgBase`

Locally I can reach 78% code coverage with the change of this PR but on Travis it should a bit better (between 80% and 85%, because of the control_node_serial source coverage).

Since this PR is touching existing open nodes, it should be first fully tested with integration tests, on all CI nodes.

### DIFF
--- a/gateway_code/control_nodes/cn_no/tests/__init__.py
+++ b/gateway_code/control_nodes/cn_no/tests/__init__.py
@@ -1,0 +1,18 @@
+# This file is a part of IoT-LAB gateway_code
+# Copyright (C) 2015 INRIA (Contact: admin@iot-lab.info)
+# Contributor(s) : see AUTHORS file
+#
+# This software is governed by the CeCILL license under French law
+# and abiding by the rules of distribution of free software.  You can  use,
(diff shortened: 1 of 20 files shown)


aabadie:
Integration tests are passing smoothly and Travis CI is green. Let's merge!

PR #76, opened 2018-09-26
open_nodes: use generic CMSIS DAP serial port for samr21

aabadie:
Apparently the debug interface of the SAMR21 is also something generic, like DAPLink, JLink of STLink.  For this board, it's CMSIS-DAP.

This PR updates the Edbg open node with a new TTY path and also change the associated udev rule.


aabadie:
Tested on CI nodes (arduino-zero) and works. Let's merge.

PR #82, opened 2018-10-18
utils/openocd_cmd: add support for all available open nodes

aabadie:
This PR allows to easily flash any kind of open nodes that support openocd from a gateway.

### DIFF
--- a/gateway_code/nodes.py
+++ b/gateway_code/nodes.py
@@ -116,10 +116,7 @@ def import_all_nodes(pkg_dir):
     for (module_loader, name, _) in pkgutil.iter_modules([pkg_dir]):
         if name in ['tests', 'common']:
             continue
-        try:
-            module_loader.find_module(name).load_module(name)
-        except ImportError:
-            pass
+        module_loader.find_module(name).load_module(name)
 
 
 import_all_nodes('open_nodes')
--- a/gateway_code/utils/cli/avrdude.py
+++ b/gateway_code/utils/cli/avrdude.py
@@ -28,35 +28,30 @@ Usage:
 
 """
 import argparse
-from .. import avrdude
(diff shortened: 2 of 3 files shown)


rienafairefr:
You could use the node plugin registry instead of hardcoding the nodes? Then you can check if the class is subtype of `NodeOpenOcd`.

aabadie:
> You could use the node plugin registry instead of hardcoding the nodes? Then you can check if the class is subtype of NodeOpenOcd.

Thinking of it and I like the idea. I'll give it a try.

aabadie:
@rienafairefr, I updated the PR like you suggested also applied the same logic for avrdude based nodes.
I tested with an M3 and ST Lrwan1 board and it works fine. Even without the try/except in the automatic nodes module importer logic, so I also removed it!

gateway_code/nodes.py
```diff
@@ -116,10 +116,7 @@ def import_all_nodes(pkg_dir):
     for (module_loader, name, _) in pkgutil.iter_modules([pkg_dir]):
         if name in ['tests', 'common']:
             continue
-        try:
-            module_loader.find_module(name).load_module(name)
-        except ImportError:
-            pass
+        module_loader.find_module(name).load_module(name)
```
rienafairefr, line 119: weird, but good, that it now works :-)

rienafairefr:
Did you test on an avr based board as well ?

aabadie:
> Did you test on an avr based board as well ?

No. It's possible on one of the CI node (there's a zigduino on Strasbourg and a Leonardo in Grenoble).

aabadie:
@rienafairefr, I tested on leonardo and zigduino and it also works. I think we can merge :)

PR #89, opened 2018-11-29
Tools

rienafairefr:
re-reading the code because I was trying to fix firefly flashing bug I found the amount of time we did a dance around subprocess.Popen/call/output/stdout quite high, especially when we have many flashing tools, and they all do the same thing. 

- This PR creates a `Tool` class from which all derive (OpenOcd, Edbg, Avrdude, CC2538)
- This `Tool` class uses subprocess.check_output instead of subprocess.call to get the stdout/stderr of the subprocess, and display it in some cases: for all cases, when verbose is True, or only in case the tool returns an error (return code !=0). 

AFAIK, tests were green (probably still some linting |-) ). But I'm still in the process of validating that this doesn't break integration tests for all the boards.

4a6bfb6 is unrelated, the usual discussion with --cov in tox.ini instead of setup.cfg

### DIFF
--- a/gateway_code/utils/tools.py
+++ b/gateway_code/utils/tools.py
@@ -24,6 +24,7 @@
 import logging
 import os
 import shlex
+import tempfile
 
 from gateway_code.utils import subprocess_timeout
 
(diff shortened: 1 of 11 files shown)


aabadie:
Nice one! It factorizes useful things.

A few comments though:
- Tool is too generic, I think FlashTool is better.
- The commit about --cov could go in its own PR, or be merged directly to master
- Still nitpicky but commit messages can also be improved: 
Example: `check_output to get subprocess's output, to print in case of error, or print always with verb=True` could be changed to a multi-line message:
```
utils: use check_output to get subprocess's output

It allows to print stdout and stderr in case of error, or always print when verb=True
```
This makes the display in GitHub nicer as well.

gateway_code/utils/tools.py
```diff
@@ -0,0 +1,70 @@
...
+        return {'args': args, 'stderr': subprocess_timeout.STDOUT}
+
+    def call_cmd(self, command_str):
+        """ Run the given command_str."""
+
+        kwargs = self.args(command_str)
+        LOGGER.info(kwargs)
+
+        try:
+            output = subprocess_timeout.check_output(timeout=self.timeout, **kwargs)
```
aabadie: You don't necessarily need check_output to get the stdout/stderr of the process. The previous `call` has named parameters for this and they can be used the same. And the change will be minimal in the tests.

gateway_code/utils/tools.py
```diff
@@ -0,0 +1,70 @@
...
+    def __init__(self, verb=False, timeout=TIMEOUT):
+        self.verb = verb
+        self.timeout = timeout
+        self.name = self.__class__.__name__
+
+    def args(self, command_str): # pylint:disable=no-self-use
+        """splits a command into arguments for Popen"""
+
+        args = shlex.split(command_str)
+        return {'args': args, 'stderr': subprocess_timeout.STDOUT}
```
aabadie: The other problem with this is that the interactive output of the tool is lost. Example: in the integration tests, when testing a debug session with openocd, the output is displayed in real time by the process.
Getting stderr/stdout is already possible in the current state.

aabadie:
>  I think FlashTool is better.

Well, now I changed my mind about this and I don't see the real benefit of having a parent class. The only thing it brings is a common `call_cmd` which takes the cmd string as input, the parsing is done internally. So this could just be a wrapper function in a separate module, maybe `subprocess_utils.py` or something like this ?

gateway_code/utils/tools.py
rienafairefr: with subprocess.call, we can have stdout, if it's a file, we can only open a temporary file in /tmp (I tried with file-like objects but it's not designed for that), that's the approach you'd take ?

rienafairefr:
Yes, I'm only putting call_cmd in the parent class, because that's the code that was repeated 4 times in the different tools. Also the `out`, `verb`, `timeout` common attributes, now in a common constructor. That could be a function, sure, but then each tool needs to handle the verb/timeout attribute.

gateway_code/utils/tools.py
rienafairefr: yeah, I think in `subprocess.Popen(**self.args(cmd_debug))` for openocd we shouldn't use the common `args` method, or rewrite the stdout parameter before the Popen so that the interactive output is not lost, thanks

rienafairefr:
@aabadie I've removed the need for check_output, works fine using a temporary file. I've unit tested the logic for the logger output.
Some integration tests fail because nodes have low space on disk, also getting NFS "Stale file handle" while logging in the `tools.py`, I'm investigating.

rienafairefr:
@aabadie there is some weird gosu keyserver thing failing the build, related discussion https://github.com/tianon/gosu/issues/39 , trying fallback mechanism right now, but I'll put that on separate PR

PR #90, opened 2018-12-13
fix firefly Flash update

rienafairefr:
as discussed, using the Zigduino model (stopping the serial redirection when flashing), it seems to work now :+1: 
Once merged I'll deploy the yocto image with the changes onto prod nodes ASAP.

### DIFF
--- a/gateway_code/open_nodes/node_firefly.py
+++ b/gateway_code/open_nodes/node_firefly.py
@@ -72,7 +72,7 @@ class NodeFirefly(OpenNodeBase):
 
         common.wait_no_tty(self.TTY)
         ret_val += common.wait_tty(self.TTY, LOGGER)
-        ret_val += self.flash(firmware_path)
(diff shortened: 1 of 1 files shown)


aabadie:
If everything goes like it should, I predict that the unit tests won't pass ;)

aabadie:
But, in any case that's super nice that you have a fix for this :)

aabadie:
hmm, it worked.

rienafairefr:
If you want peace of mind of the test on CI node passing, I still need to understand why the /var/volatile fills up (nodes.test_gateway_code doesn't pass on my firefly in ci)

aabadie:
maybe because your yocto image is quite old and the setup install some dependencies via pip. In newer images, all python dependencies are already provided by the yocto image.

rienafairefr:
good idea. I'll check that

PR #92, opened 2018-12-18
open_nodes: add verify method in OpenNodeBase class

fsaintma:
This PR is a preleminary work to integrate RPI3 Open Linux node. Indeed we add verify method in the OpenNodeBase class instead of a global method which verify when you load open nodes class from the registry.
It provides to override this method in open nodes class (ex: NodeA8) and adapt the verification. Thus in Linux open nodes class we won't check attributes like available autotests or firmware target (e.g. ELF_TARGET)

### DIFF
--- a/gateway_code/nodes.py
+++ b/gateway_code/nodes.py
@@ -19,6 +19,8 @@
 # The fact that you are presently reading this means that you have had
 # knowledge of the CeCILL license and that you accept its terms.
 
+# pylint: disable=no-member
+
 """ Common logic for plugin nodes classes """
 import abc
 import os
--- a/gateway_code/nodes.py
+++ b/gateway_code/nodes.py
@@ -115,20 +115,21 @@ class OpenNodeBase(object):
         """ Verifiy the open node """
         ret_val = 0
         # Tuple with (class, machine) run 'elftarget.py' on a node firmware
-        if getattr(cls, 'ELF_TARGET', None) is None or \
-           len(cls.ELF_TARGET) != 2:
+        if (getattr(cls, 'ELF_TARGET', None) is None or
+                len(cls.ELF_TARGET) != 2):
             ret_val += 1
 
         for firmware_attr in ('FW_IDLE', 'FW_AUTOTEST'):
             firmware = getattr(cls, firmware_attr, None)
-            if firmware and \
-               not elftarget.is_compatible_with_node(firmware, cls):
+            if (firmware is not None and not
+                    elftarget.is_compatible_with_node(firmware, cls)):
                 ret_val += 1
 
         required_autotest = {'echo', 'get_time'}  # mandatory
-        if getattr(cls, 'AUTOTEST_AVAILABLE', None) is None or \
-           not required_autotest.issubset(cls.AUTOTEST_AVAILABLE):
+        if (getattr(cls, 'AUTOTEST_AVAILABLE', None) is None or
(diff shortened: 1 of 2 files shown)


gateway_code/open_nodes/tests/open_nodes_test.py
```diff
@@ -79,6 +79,22 @@ class MyNode(OpenNodeBase):
         ELF_TARGET = ('ELFCLASS32', 'EM_ARM')
         AUTOTEST_AVAILABLE = ['echo', 'get_time']
 
+        def setup(self, firmware_path):
```
aabadie: I don't think setup/teardown/status are needed. Just `verify`.

gateway_code/nodes.py
```diff
@@ -108,6 +108,23 @@ def status(self):
         """ Status of the node """
         pass  # pragma: no cover
 
+    def verify(self):
+        """ Verifiy the open node """
+        ret_val = 0
+        # Tuple with (class, machine) run 'elftarget.py' on a node firmware
+        if len(self.ELF_TARGET) != 2:
+            ret_val += 1
```
aabadie, line 120: This line is not covered by tests. It is possible to add a custom node class in open_nodes_test module to trigger and verify the correct behaviour of this method.

This is the same for line 121 and 125.

aabadie (commented):
If you want to check locally the lint errors, just use `tox -e code_check` because otherwise, since you pushed your branch directly on the main repository, 2 runs are triggered by a single push (one for the PR, one for the branch).

gateway_code/open_nodes/tests/open_nodes_test.py
fsaintma: no there is an error if you not put all abstract methods. I check it

aabadie: Ok, I had a look the previous code, and the check of `verify` should be done in another place. See https://github.com/iot-lab/iot-lab-gateway/blob/master/gateway_code/open_nodes/tests/open_nodes_test.py#L137

You can also subclass this class to check error cases.

Does it work when you leave the test_registry_open_node() unchanged ?

fsaintma: Hi I will add an assert board_instance.verify() == 0 at the end of the file open_nodes_test.py
It doesn't work with test_registry_open_node() unchanged:
Error message: Can't instantiate abstract class MyNode with abstract methods setup, status, teardown

aabadie: It is not normal because this function doesn't instanciate MyNode. It just checks that the class is correctly registered.

gateway_code/nodes.py
```diff
@@ -137,25 +156,12 @@ def _node_class(cld, board_type):
...
-    assert required_autotest.issubset(output_class.AUTOTEST_AVAILABLE)
-
-
 def open_node_class(board_type):
     """Return the open node class implementation for `board_type`.
 
     :raises ValueError: if board class can't be found """
     output_class = _node_class(OpenNodeBase, board_type)
-    _verify_open_node_class(output_class)
+    assert output_class().verify() == 0
```
aabadie: Your problem is here.

aabadie:
I pushed an update that fixes the problem with test coverage. Let's see how it goes on Travis

PR #93, opened 2018-12-19
Linux open node

fsaintma:
Add linux_open_node_type config to learn which open node is on the Linux node (e.g. A8 or RPI3)
   * A8 Linux node <-> A8-M3 open node
   * RPI3 Linux node <-> SAMR21|FIREFLY|... open node 

Update open A8 node class with verify method (e.g. no check) and add A8-M3 open node class with corresponding attributes (eg. autotest available, elf target, ...)
Update autotest to detect Linux node and launch autotest on the Linux open node. Thus adapt code to provide a generic Linux ssh interface access which start serial redirection on the Linux node.


bin/rules.d/a8-m3.rules
```diff
@@ -1,4 +1,4 @@
 # Open A8 M3 node
-SUBSYSTEM=="tty", SUBSYSTEMS=="usb", ENV{ID_SERIAL}=="IoT-LAB_A8-M3",         ENV{ID_USB_INTERFACE_NUM}=="01",  SYMLINK+="ttyA8_M3 iotlab/ttyA8_M3"
+SUBSYSTEM=="tty", SUBSYSTEMS=="usb", ENV{ID_SERIAL}=="IoT-LAB_A8-M3",         ENV{ID_USB_INTERFACE_NUM}=="01",  SYMLINK+="ttyA8_M3"
```
aabadie: I think there is no need for changing this. I would even think it will useful in the future with custom open node behind the Linux node. Because the TTY will always be `/dev/iotlab/tty<something>`

PR #94, opened 2018-12-21
open_nodes: introduce node with no programmable node and rtl_sdr node

aabadie:
This PR adds the support of the RTL-SDR dongle as a regular open node.
- this node provides no programmable node => it was necessary to have a base open node with no node
- this node manage the RTL-SDR power and rtl_tcp daemon. The power of the dongle is controlled via a YepKit module (default model with 3 USB ports, the dongle must be plugged on port 3).

This code can be tested on devsaclay.

### DIFF
--- a/gateway_code/open_nodes/node_rtl_sdr.py
+++ b/gateway_code/open_nodes/node_rtl_sdr.py
@@ -25,8 +25,6 @@ import logging
 import shlex
 import subprocess
 
-import gateway_code.utils.ftdi_check
-
 from gateway_code.common import logger_call
 from gateway_code.utils.rtl_tcp import RtlTcp
 from gateway_code.open_nodes.common.node_no import NodeNoBase
--- a/gateway_code/open_nodes/node_rtl_sdr.py
+++ b/gateway_code/open_nodes/node_rtl_sdr.py
@@ -1,0 +1,65 @@
+# -*- coding:utf-8 -*-
+
+# This file is a part of IoT-LAB gateway_code
+# Copyright (C) 2015 INRIA (Contact: admin@iot-lab.info)
+# Contributor(s) : see AUTHORS file
+#
+# This software is governed by the CeCILL license under French law
+# and abiding by the rules of distribution of free software.  You can  use,
+# modify and/ or redistribute the software under the terms of the CeCILL
+# license as circulated by CEA, CNRS and INRIA at the following URL
+# http://www.cecill.info.
+#
+# As a counterpart to the access to the source code and  rights to copy,
+# modify and redistribute granted by the license, users are provided only
+# with a limited warranty  and the software's author,  the holder of the
+# economic rights,  and the successive licensors  have only  limited
+# liability.
+#
+# The fact that you are presently reading this means that you have had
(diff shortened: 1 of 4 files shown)


gateway_code/open_nodes/node_rtl_sdr.py
```diff
@@ -0,0 +1,73 @@
...
+        kwargs = {'args': shlex.split(command_str)}
+        try:
+            return subprocess_timeout.call(**kwargs)
+        except subprocess_timeout.TimeoutExpired as exc:
+            LOGGER.error("Command '%s' timeout: %s", command_str, exc)
+            return 1
+
+    @logger_call("Node No: Setup of no node")
+    def setup(self, firmware_path=None):
+        """Does nothing."""
```
rienafairefr: copy-pasted doc string, this method is not doing nothing ;-)

gateway_code/open_nodes/node_rtl_sdr.py
```diff
@@ -0,0 +1,73 @@
...
+# Yepkit module.
+YKUSHCMD = "sudo ykushcmd {model} {cmd} {port}"
+
+
+class NodeRtlSdr(NodeNoBase):
+    """ Open node Rtl SDR implementation """
+
+    TYPE = 'rtl_sdr'
+    RTL_TCP_PORT = 50000
+    RTL_TCP_FREQ = 868000000
```
rienafairefr, line 44: hardcoded to 868 MHz I see, wouldn't it make more sense to have this be configurable at experiment submit by the user ? Or maybe we can send something on port 50000 to change it once the experiment is started ?

gateway_code/open_nodes/node_rtl_sdr.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def __init__(self):
+        self.rtl_tcp = RtlTcp(self.RTL_TCP_PORT, self.RTL_TCP_FREQ)
+
+    @staticmethod
+    def _call_cmd(command_str):
```
rienafairefr: this method is really starting to be repeated every where, don't you think ?

gateway_code/open_nodes/node_rtl_sdr.py
aabadie: This default value is selected for convenience but in the end, it doesn't make any difference: in GNU Radio, the configuration box change it to what the user want. I expect people to use a tool like GNU Radio.

gateway_code/open_nodes/node_rtl_sdr.py
aabadie: I agree. And I think I'll need it again very soon but #89 won't be adapted for that. There's no need for a base class just for that. I'll open a PR (an alternative to #89)

aabadie: see #98

aabadie: Now that I'm thinking at this a second time, I think reusing `subprocess_timeout` is useless in this case. Just a regular `subprocess.call` will be enough, in try/except to catch errors. The underlying command is not supposed to take time. I'll change that.

PR #96, opened 2019-01-04
tests: fix assertEquals/assertNotEquals deprecated-method error with pylint

aabadie:
The corresponding methods are assertEqual/assertNotEqual.

I think I get this because I updated my system to ubuntu cosmic which comes with a new python 2.7 version. It doesn't seem to be a bad idea to update. Also tested directly on a gateway and works there as well.

The update was performed semi-automatically using:
```
$ git grep -l 'self.assertEquals(' | xargs sed -i 's/self.assertEquals(/self.assertEqual(/g'
$ git grep -l 'self.assertNotEquals(' | xargs sed -i 's/self.assertNotEquals(/self.assertNotEqual(/g'
```


aabadie:
So apparently this is a general problem, see #95

PR #97, opened 2019-01-08
fallback key servers for gosu

rienafairefr:
this is for the Docker testing system: this fixes the problem of the gosu GPG key not being found sometimes, by using multiple fallback servers, seen in a build of #89


aabadie:
Something was wrong with Travis service integration and no CI was run. I updated the config. Can you update your branch to trigger a build ?

aabadie:
Otherwise it looks good, I just want to make sure that the CI is ok before merging.

PR #99, opened 2019-01-09
CC2650 launchpad

rienafairefr:
WIP support for the CC2650 launchpad

- flashing is done through a TI modified openOCD
- there is a firmware verification step because if the user overwrites a certain configuration space then the board would be bricked or at least requiring manual intervention

It looks like there is still some issues with the serial link, autotests are failing, but it's getting good for the flashing part. Will test in real iot-lab gateway soon


aabadie:
I reworked a bit this PR:
- removed the firmware verifier, it can be added later but not in this PR
- removed the change related to CMSIS_DAP udev rules (unrelated)
- changed the location of the TO version of openocd to `/opt/openocd-ti`. The related PR in iot-lab-yocto has to be updated as well (will do it later).
- added a README that explains how to setup the simplelink programmer (update the XDS110 firmware using CCS7)
- rebased this PR on top of #127 and #128 

Unfortunately, I'm not able to make it work within Docker and I don't understand why: when I flash on my machine (not from Docker), using the same openocd version, same output, same autotest firmware, everything works. But from Docker it doesn't: the firmware doesn't work for some reason.
I'll plug one board in devsaclay and see if it works when plugged on a gateway. If it does, I'll merge.

PR #100, opened 2019-01-17
Programmer

fsaintma:
Replace old programmer scripts (openocd_cmd, arvrdude_cmd, ...) with one generic programmer.
We autodetect the open|control node with board config and execute the corresponding command (flash, reset, debug)
The idea is to provide these scripts on the A8|RPI3 Yocto Linux image and to replace all old programmer scripts (flash_a8_m3, flash_cn, flash_m3). Thus it will be possible to execute the autotest integration tests 
on the Linux open nodes like RPI3 node and flash the good firmware.


gateway_code/utils/cli/programmer.py
```diff
@@ -0,0 +1,113 @@
...
+def reset():
+    """ reset node function """
+    parser = _setup_parser(reset.__name__)
+    opts = parser.parse_args()
+    node = _get_node(opts.cn)
+    if hasattr(node, reset.__name__):
+        ret = node.reset()
+        _print_result(ret, reset.__name__)
+        return ret
+    raise ValueError(CMD_ERROR.format(reset.__name__, node.TYPE))
```
aabadie: Instead of raising an exception (which is not trapped), just return a value != 1 and print something to stdout ("Reset not supported for board xxxx"). Same for debug and flash.

aabadie: or stderr

PR #101, opened 2019-01-17
Add RPI3 Linux open node

fsaintma:


aabadie:
Let's go with this one !!!

PR #102, opened 2019-02-11
Serial redirection CLI

fsaintma:
*  add open node TTY, baudrate autodetection


aabadie:
Looks good and no particular comments. Let's merge!

PR #104, opened 2019-02-18
add basic support for pycom open nodes (Micropython inside)

aabadie:
This PR provides a POC for supporting [Pycom](https://pycom.io/) nodes in IoT-LAB:
- only the MicroPython REPL is available via the serial_redirection
- the serial_redirection is slightly adapted by this PR to allow passing extra socat options: indeed the Micropython REPL requires `crnl` which is not used with the regular serial redirection.
- flashing the underlying firmware or uploading a Python script is not possible
- basic interaction with the REPL is possible via the web terminal (no copy/paste which is annoying)
- access to the REPL is possible **from the frontend** but not using `nc`. Users have to use the following commands:
  ```
  $ stty -icanon
  $ socat -,echo=0 tcp:<pycom node>:20000
  ```
- it should be possible to access the REPL via `pymakr` (Atom extension for micropython) using ssh tunneling + socat but I haven't tried yet.

There's one node for testing in devsaclay (pycom-1). The type of pycom board is provided by the `archi` field in the resources file (`pycom:fipy`, `pycom:lopy`, etc). The one in devsaclay is a `fipy`.

iotlab-micropython

### DIFF
--- a/bin/rules.d/pycom.rules
+++ b/bin/rules.d/pycom.rules
@@ -1,3 +1,9 @@
 # open nodes TTY
-SUBSYSTEM=="tty", SUBSYSTEMS=="usb", ENV{ID_USB_INTERFACE_NUM}=="00", SYMLINK+="iotlab/ttyON_PYCOM"
-SUBSYSTEM=="usb", ATTR{idProduct}=="6015", ATTR{idVendor}=="0403", MODE="0664", GROUP="dialout"
+
+# pytrack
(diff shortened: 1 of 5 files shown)


rienafairefr:
I think we should keep the` board_type:radio` standard, would be more conforming with `pycom_fipy:multi` , maybe `pycom|fipy:multi` or another delimiter, or simply `fipy` (will you even deploy multiple different types of pycom board ?)

aabadie:
I need this PR and will merge it now.

PR #112, opened 2019-07-02
Support for flashing binary files with openocd nodes

rienafairefr:
sister PRs:
[testbed-webportal #90](https://github.com/iot-lab/testbed-webportal/pull/90)
[testbed-manager #74](https://github.com/iot-lab/testbed-manager/pull/74)
[site-manager #10](https://github.com/iot-lab/site-manager/pull/10)
<-

### DIFF
--- a/gateway_code/utils/openocd.py
+++ b/gateway_code/utils/openocd.py
@@ -60,6 +60,11 @@ class OpenOCD(object):
              ' -c "reset run"'
              ' -c "shutdown"')
 
+    FLASH_BIN = (' -c "reset halt"'
+                 ' -c "reset init"'
+                 ' -c "program {0} verify {1}"'
+                 ' -c "shutdown"')
+
     DEBUG = ' -c "reset halt"'
     TIMEOUT = 100
 
@@ -98,11 +103,13 @@ class OpenOCD(object):
         """ Reset """
         return self._call_cmd(self.RESET)
 
-    def flash(self, elf_file):
+    def flash(self, fw_file, binary=False, offset=None):
         """ Flash firmware """
         try:
-            elf_path = common.abspath(elf_file)
-            return self._call_cmd(self.FLASH.format(elf_path))
+            path = common.abspath(fw_file)
+            if binary:
+                return self._call_cmd(self.FLASH_BIN.format(path, offset))
+            return self._call_cmd(self.FLASH.format(path))
         except IOError as err:
             LOGGER.error('%s', err)
             return 1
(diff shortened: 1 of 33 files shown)


gateway_code/utils/openocd.py
```diff
@@ -60,11 +60,16 @@ class OpenOCD(object):
...
+                 ' -c "reset init"'
+                 ' -c "program {0} verify {1}"'
+                 ' -c "shutdown"')
+
     DEBUG = ' -c "reset halt"'
     TIMEOUT = 100
 
     def __init__(self, openocd_args,
-                 verb=False, timeout=TIMEOUT):
+                 verb=True, timeout=TIMEOUT):
```
aabadie: Maybe set this back to False ?

gateway_code/utils/openocd.py
```diff
@@ -98,11 +103,14 @@ def reset(self):
         """ Reset """
         return self._call_cmd(self.RESET)
 
-    def flash(self, elf_file):
+    def flash(self, file, binary=False, offset=None):
         """ Flash firmware """
         try:
-            elf_path = common.abspath(elf_file)
-            return self._call_cmd(self.FLASH.format(elf_path))
+            path = common.abspath(file)
```
aabadie: `file` is a Python keyword, maybe use `fw_file` instead ?

aabadie:
There are some PEP8 reported by the CI + some unit tests also require an update.

rienafairefr:
@aabadie OK, I've integrated the reviews, and will fix the pep8 and unit tests :-)

aabadie:
FYI, I fixed the remaining failing unit tests and added new ones (there were some new uncovered lines). I went for using return codes != 0 instead of raising exception as it's done this way already.
We need to perform integration tests on related boards before merging this. If it works, it can be merged since it's transparent for the rest of the intrastructure IIUC.

aabadie:
I did more tests of this feature and after updating a few things, I think it can be merged now.

- Integrations tests are passing
- I manually tested the feature on the dev platform (nrf52dk in devsaclay, samr21 in devgrenoble) and it works
- Now edbg flasher also supports the binary mode (and offset). This is required with samr21 boards.

aabadie:
Some more testing on different boards showed that ST MCU ROM is not starting at the address 0 and require to also pass the start address of the ROM. That's what 9e99149 do. I tested it on st-lrwan1 boards and it works. Currently I tested this PR on nrf52dk, nrf52840dk, samr21, arduino-zero and st-lrwan1. And all are working.

PR #113, opened 2019-07-08
support for nucleo-f070rb

rienafairefr:
These are the nodes at nantes, 
board STM32F0xx
nucleo 64,
the firmware was built via ci-firmwares without any issues. Still doing some integration tests


aabadie:
Which F0 exactly ? (f091, f072, f070, f030 ?)
You can have a look at the board names used in RIOT to get an idea.

rienafairefr:
F070RB, you think the board name should include more of the whole name ?

rienafairefr:
I built in RIOT using `BOARD=nucleo-f070`

rienafairefr:
OK, pushed with modified name, `st_nuclo64f070`

aabadie (changes requested):
It's so simple to add new boards now :)

Only one comment: I would simply rename the board in `nucleo-f070rb`. This is the commercial name of this plateform, see [here](https://www.st.com/content/st_com/en/products/evaluation-tools/product-evaluation-tools/mcu-mpu-eval-tools/stm32-mcu-mpu-eval-tools/stm32-nucleo-boards/nucleo-f070rb.html)

rienafairefr:
@aabadie OK I just renamed the board to `nucleo_f070rb`

aabadie (approved):
All green, let's go!

PR #114, opened 2019-07-10
support for samd21 Xplained pro board

spillement:
Bonjour super ingé IoT_Lab,
proposition des fichiers supports pour les cartes SAMD21 Xplained Pro qui sont en train d'être déployées à Nantes.

Fait avec Matthieu à coté, et heureusement qu'il était là ;-) On pourra échanger sur la doc disponible sur le gitHub.


rienafairefr:
Some info to add:
- we built using [iot-lab/ci-firmwares](https://github.com/iot-lab/ci-firmwares), using the RIOT BOARD `samd21-xpro`, and using `master` branch of RIOT
- normally the `get_uid`, `leds_***` autotests should have worked fine, just like for samr21, but they failed for some reason, which might be fixed by a `DIRTY_SERIAL`, I didn't have time to investigate more and just deactivated those autotests because I'm a lazy guy ^^. 

we deployed `samd21-1 - samd21-5` with `at86rf233` radio extensions on devnantes, which are not really booting/accessible for now, because we're still battling with connectivity from the server (in Univ-Nantes' datacenter) to the switch (in a Polytech Nantes room)

As Sébastien said, the documentation for developing support for new node is a bit lacking. We also had to battle with the SI rules (which block ports 80 and 443, you have to use their proxy server), and deeply adapt the `Dockerfile` and `Makefile` to handle that proxy thing.

PR #117, opened 2019-08-27
open_node/st-lrwan1: enable dual bank support with openocd

aabadie:
This boards provide a dual bank 192kB flash memory and the current OpenOCD configuration only works with single bank which means only 96kB of flash is available.

This PR is based on #112 because I fear there will be conflicts with the ROM_START_ADDR attribute introduced there.


aabadie:
Also note that this was tested on devsaclay and it works like a charm :)

PR #118, opened 2019-08-27
open_nodes/iot-lab: refactor openocd script + unify fox/m3/a8_m3

### LINKED ISSUE #108: Review new openocd scripts using 'ftdi' interface instead of ft2232
http://openocd.zylin.com/2415

aabadie:
This PR is refactoring the OpenOCD configuration used with iotlab boards (m3, fox, a8-m3) and is enabling the use of recent openocd version.

Since the same OpenOCD configuration is used for each of these boards, only one is kept and all now use the same.

This needs careful testing but so far I had no issue (even after reflashing several times an m3). Note the OpenOCD config was taken from RIOT config file and is known to work.
This PR could be useful if we decide to allow flashing binary firmwares on M3 (see #112).

Another consequence is that we could get rid of the legacy openocd 0.9 version :) and then remove it from the Yocto images, yeah!

fixes #108


aabadie:
fixes #108

aabadie:
Discussed IRL with @cladmi about this and the concept is ok. The control node script should also be adapted.

But the removed command `ft2232_device_desc "A8-M3"` should be kept if possible. I'll test that.

cladmi:
I never used the `ftdi` openocd driver before as it was not working properly in RIOT, and was thinking that it was done properly there, so that the issue was with openocd/ftdi/handling of our hardware.

But, the file was just completely wrong in RIOT.

aabadie:
Thanks @cladmi for the comment which goes in the direction of this PR.

I just pushed an update for the control nodes, which I initially forgot. With this, the use of openocd 0.9 becomes completely useless.

gateway_code/static/iot-lab.cfg
```diff
@@ -0,0 +1,4 @@
+source [find interface/ftdi/iotlab-usb.cfg]
+source [find target/stm32f1x.cfg]
+
+reset_config trst_and_srst connect_assert_srst
```
cladmi, line 6: Maybe keep the comment about "TRST/SRST" it's not the most trivial thing.

gateway_code/open_nodes/node_fox.py
```diff
@@ -33,9 +33,9 @@ class NodeFox(NodeOpenOCDBase):
     TYPE = 'fox'
     TTY = '/dev/iotlab/ttyON_FOX'
     BAUDRATE = 500000
-    OPENOCD_PATH = 'openocd'  # Use old openocd
-    OPENOCD_CFG_FILE = static_path('iot-lab-fox.cfg')
-    OPENOCD_OPTS = ('target/stm32f1x.cfg',)
+    OPENOCD_PATH = '/opt/openocd-dev/bin/openocd'
+    OPENOCD_CFG_FILE = static_path('iot-lab.cfg')
+    OPENOCD_OPTS = (static_path('iot-lab-fox.cfg'),)
```
cladmi, line 38: Was strange that 'opts' was the more generic one before, so flipping it now makes really sense.

gateway_code/static/iot-lab-cn-m3.cfg
```diff
@@ -1,14 +1 @@
...
-ft2232_device_desc "ControlNode"
-ft2232_vid_pid 0x0403 0x6010
-
-jtag_nsrst_delay 100
-jtag_ntrst_delay 100
-
-# use combined on interfaces or targets that can't set TRST/SRST separately
-reset_config trst_and_srst
-
+ftdi_device_desc "ControlNode"
```
cladmi, line 1: It took me a few seconds for that one. But I re-read the file name and understood :D

gateway_code/static/iot-lab-cn.cfg
```diff
@@ -1,14 +1,2 @@
...
-ft2232_vid_pid 0x0403 0x6011
-
-jtag_nsrst_delay 100
-jtag_ntrst_delay 100
-
-# use combined on interfaces or targets that can't set TRST/SRST separately
-reset_config trst_and_srst
-
+ftdi_device_desc "ControlNode"
+ftdi_vid_pid 0x0403 0x6011
```
cladmi, line 5: Maybe comment here, that it must overwrite the default value given in `openocd` official `iotlab-usb.cfg`

cladmi (commented):
I would keep a bit more comments but otherwise seems good.

gateway_code/static/iot-lab-cn.cfg
cladmi: The low level reason, it is that the M3 and A8-M3 have a 2 ports `FT2232 ` where the control node has a 4 port `FT4232`.

aabadie:
Thanks for having a second look @cladmi. Your comments are now addressed. Before merging, I'll test this change on an experiment in IoT-LAB: flash the same firmware 1000 times in a loop. Just to see if it can never fail.

cladmi:
You can also test some after open node poweroff/on if you want to add other constraints.

aabadie:
I could re-flash almost 900 times a firmware on an M3 and FOX without issues. I was also doing a stop/start on the FOX between each reflash.

aabadie:
Tested on an A8: iotlab_flash/reset/debug works.

PR #122, opened 2020-04-30
Add Python 3 compatibility

aabadie:
- Unit tests are now working on Python 2.7, 3.6 and 3.8 (all tested)
- Lots of lints fixed

In the current state, the integration tests are failing (Python 2) because of commit 5a2df5b. To be able to test with Python 3, it would be better to update the Python packages installed in the Yocto image and deploy this image on one of the ci gateways.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ PORT ?= 8080
 SERIAL_PORT ?= 20000
 CONTROL_NODE_TYPE ?= no
 POSARGS ?=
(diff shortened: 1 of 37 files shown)


aabadie:
Integration tests are successful on M3, A8, st-lrwan1 and samr21.

There's still one issue: Python 3 is not tested on Travis because there's no Python 3 interpreter in the Docker container. We should fix this before merging.

PR #131, opened 2020-07-08
open_nodes: add support for STM32 p-nucleo-wb55

aabadie:


aabadie:
Tested on devsaclay and in Docker: all works.

PR #137, opened 2021-04-12
fix Python3 support with encoding

fsaintma:
Fix all encoding integration & autotests bugs with Python3 support.

Integration tests (fab nodes.test_gateway_code task) pass on Devgrenoble site (m3-17 & a8-17) with Yocto A8 target and:
* Yocto Krogoth version image
* [Yocto Dunfell version image](https://github.com/iot-lab/iot-lab-yocto/pull/39)

There is still Yocto RPI3 target support to be tested

### DIFF
--- a/gateway_code/autotest/autotest.py
+++ b/gateway_code/autotest/autotest.py
@@ -115,8 +115,8 @@ class AutoTestManager(object):
     @staticmethod
     def get_local_mac_addr():
         """ Get eth0 mac address """
-        mac_addr = check_output(MAC_CMD, stderr=STDOUT, shell=True).strip()
-        return mac_addr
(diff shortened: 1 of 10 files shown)


aabadie:
Is there a reason to not take into account #135 ?

fsaintma:
Hi @aabadie. I think that your PR mix two things: remove Python2 version and fix Python3 support (encoding). I prefer do a PR which fix all Python3 encoding problems (with integration tests) and continues to work with Python3 (Yocto Dunfell) and Python2 (Yocto Krogoth) version. So I prefer release a new gateway code version (2.10.1) with this PR and then use a part of #135  to only remove Python2 support in the code (2.11.0). Are you agree ?

aabadie:
That sounds reasonable :)

PR #139, opened 2021-09-21
Zigduino autotests fix

schrein:
The board seems to slow for only one second to setup a connection to open node during autotests. Empiric tests show a value a 3 seconds is successful on IoT-LAB legacy gateway and Raspberry Pi 2 gateway.

### DIFF
--- a/gateway_code/autotest/autotest.py
+++ b/gateway_code/autotest/autotest.py
@@ -99,6 +99,7 @@ class AutoTestManager(object):
         self.g_m = gateway_manager
         board_cfg = board_config.BoardConfig()
         self.on_class = board_cfg.board_class
+        self.on_board_type = board_cfg.board_type
         self.cn_class = board_cfg.cn_class
         self.linux_on_class = board_cfg.linux_on_class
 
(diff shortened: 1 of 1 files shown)


gateway_code/autotest/autotest.py
```diff
@@ -275,7 +275,7 @@ def auto_tests(self, channel=None, blink=False, flash=False, gps=False):
 
             # switch to DC and configure open node
             self._setup_open_node_connection()
-            time.sleep(1)
+            time.sleep(3)
```
aabadie: Is it possible to only use 3 when the autotests are run on an RPI2 gateway (or if the board type is Zigduino) ?
`board_type` is available with BoardConfig:
https://github.com/iot-lab/iot-lab-gateway/blob/2b2d86bd504d5a53a8e46d2f45c962712d0575e1/gateway_code/board_config.py#L32

Maybe we could add `gw_type` (but this needs to be done in the install_lib)

schrein: > Is it possible to only use 3 when the autotests are run on an RPI2 gateway (or if the board type is Zigduino) ?
> `board_type` is available with BoardConfig:

Yes good idea, I had it in https://github.com/iot-lab/iot-lab-gateway/pull/139/commits/68bc334d6c1a20e35d0dd0c0928fa6768b855aa2.

> Maybe we could add `gw_type` (but this needs to be done in the install_lib)

The autotest failed both with iotlab and rpi2 gateway. So in this case it's not useful, `board_type` conditional test works for all `gw_type`.

PR #142, opened 2021-12-20
Pycom Expansion Board

schrein:
Add support for Pycom Expansion Board (allow hard flash reset)

### DIFF
--- a/bin/rules.d/pycom.rules
+++ b/bin/rules.d/pycom.rules
@@ -7,3 +7,7 @@ SUBSYSTEM=="usb", ATTR{idProduct}=="04d8", ATTR{idVendor}=="f014", MODE="0664", 
 # pysense
 KERNEL=="ttyACM*", SUBSYSTEM=="tty", ATTRS{idVendor}=="04d8", ATTRS{idProduct}=="f012", SUBSYSTEMS=="usb", SYMLINK+="iotlab/ttyON_PYCOM"
 SUBSYSTEM=="usb", ATTR{idProduct}=="04d8", ATTR{idVendor}=="f012", MODE="0664", GROUP="dialout"
+
+# evaluation board
+KERNEL=="ttyACM*", SUBSYSTEM=="tty", ATTRS{idVendor}=="04d8", ATTRS{idProduct}=="ef98", SUBSYSTEMS=="usb", SYMLINK+="iotlab/ttyON_PYCOM"
+SUBSYSTEM=="usb", ATTR{idProduct}=="04d8", ATTR{idVendor}=="f012", MODE="0664", GROUP="dialout"
--- a/bin/rules.d/pycom.rules
+++ b/bin/rules.d/pycom.rules
@@ -8,6 +8,6 @@ SUBSYSTEM=="usb", ATTR{idProduct}=="04d8", ATTR{idVendor}=="f014", MODE="0664", 
 KERNEL=="ttyACM*", SUBSYSTEM=="tty", ATTRS{idVendor}=="04d8", ATTRS{idProduct}=="f012", SUBSYSTEMS=="usb", SYMLINK+="iotlab/ttyON_PYCOM"
 SUBSYSTEM=="usb", ATTR{idProduct}=="04d8", ATTR{idVendor}=="f012", MODE="0664", GROUP="dialout"
 
-# evaluation board
+# expansion board
 KERNEL=="ttyACM*", SUBSYSTEM=="tty", ATTRS{idVendor}=="04d8", ATTRS{idProduct}=="ef98", SUBSYSTEMS=="usb", SYMLINK+="iotlab/ttyON_PYCOM"
 SUBSYSTEM=="usb", ATTR{idProduct}=="04d8", ATTR{idVendor}=="f012", MODE="0664", GROUP="dialout"
--- a/gateway_code/open_nodes/node_pycom.py
+++ b/gateway_code/open_nodes/node_pycom.py
@@ -21,14 +21,18 @@
 
 """ Open Node Pycom experiment implementation """
 
+import os.path
 import time
 import logging
+import shlex
 import serial
 
 import gateway_code.common
 from gateway_code.common import logger_call
 from gateway_code.utils.serial_redirection import SerialRedirection
 from gateway_code.open_nodes.common.node_no import NodeNoBase
+from gateway_code import config
+from gateway_code.utils import subprocess_timeout
 
 LOGGER = logging.getLogger('gateway_code')
 PYCOM_SAFE_REBOOT_SEQUENCE = {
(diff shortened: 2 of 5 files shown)


gateway_code/open_nodes/node_pycom.py
```diff
@@ -44,6 +48,22 @@
     b"machine.reset()\r\n",
 )
 
+PYCOM_UPDATE_BIN = "sudo /usr/bin/python3 /usr/local/share/pycom/eps32/tools/fw_updater/updater.py"
+PYCOM_FLASH_ERASE_HARD = "{bin} --pic -p {port} erase_fs"
+
+LOCAL_CONFIG_DIR = '/var/local/config'
```
aabadie: This won't be compatible with the multi ON feature that will come soon. Better get the path from the environment variable, as done here:

https://github.com/iot-lab/iot-lab-gateway/blob/f9acc3fcbcd5abf1048f8cb317903a52b95aaf5d/gateway_code/config.py#L41-L43

gateway_code/open_nodes/node_pycom.py
```diff
@@ -80,7 +108,9 @@ def setup(self, firmware_path=None):
             ssh -L 20000:<pycom node>:20000 <login>@<site>.iot-lab.info
             socat PTY,link=/tmp/ttyS0,echo=0,crnl TCP:localhost:20000
         """
-        ret_val = gateway_code.common.wait_tty(self.TTY, LOGGER, timeout=10)
+        if self.expansion_board is True:
```
aabadie: Is the expansion board compatible with pytrack/pysense expansion boards ? (I think it is but unsure) Have you tried that ?

bin/rules.d/pycom.rules
```diff
@@ -7,3 +7,7 @@ SUBSYSTEM=="usb", ATTR{idProduct}=="04d8", ATTR{idVendor}=="f014", MODE="0664",
 # pysense
 KERNEL=="ttyACM*", SUBSYSTEM=="tty", ATTRS{idVendor}=="04d8", ATTRS{idProduct}=="f012", SUBSYSTEMS=="usb", SYMLINK+="iotlab/ttyON_PYCOM"
 SUBSYSTEM=="usb", ATTR{idProduct}=="04d8", ATTR{idVendor}=="f012", MODE="0664", GROUP="dialout"
+
+# evaluation board
```
aabadie: Should be expansion board I think

aabadie (commented):
There are CI failures reported and I have a few inline questions/comments.

Main question is: would it be possible to use the same hard flash reset with pysense/pytrack expansion boards ? I think they are built on the same base, just providing extra features (sensors, etc). That would avoid to check for archi and remove the special cases. That would also remove the need for a software flash reset.

gateway_code/open_nodes/node_pycom.py
schrein: Ok will improve this

gateway_code/open_nodes/node_pycom.py
schrein: Expansion bard, Pytrack and Pysense are all of them shields/daughter board. You can not plugged them together, they are exclusive. You can plugged Fipy/LoPy/Gpy on any of these shields.

bin/rules.d/pycom.rules
schrein: Yes correct, will fixed that too.

gateway_code/open_nodes/node_pycom.py
aabadie: Sure. But my question was more about the compatibility of the initial code with the expansion board. Is it so different from Pytrack/Pysense that it needs a special case ?
What kind of board are you plugging on top the expansion board ? In Saclay, Fipy are deployed for example.

schrein: We use both LoPy4 and FiPy.

For the special case, the Expansion Board 3.1 allows you to hard reset the Pycom Board at the setup (force flash erase) via the embedded PIC MCU. 

For the future,  we can also add the MicroPython firmware update support thanks to the update.py tools. Or even flash another OS on the ESP32. 

But sadly, the update.py works only with the  Expansion Board 3.1, no need to manually hard wire GPIO or push reset button like the Pysense/Pytrack v1 boards.

So I think it's good to have different level of support for pycom boards depending the base shield.

aabadie: Ok but it seems that there's an expansion board v3.0 that is compatible with pysense/pytrack. So it's more a matter of hardware version rather than a expansion board variant.
I understand the benefit of the v3.1 as you describe (can you share a good link where it's well described, I didn't find one, maybe add it to the node class docstring).

aabadie: I found [this page](https://docs.pycom.io/datasheets/expansionboards/expansion3/#differences-between-v20-and-v30) that explains the differences. Apparently the pysense/pytrack v2 are compatible to what you describe. The V3.1 doesn't provide super critical features compared to v3.0.

I'll check the versions of the expansion boards deployed in Saclay to see if they can be compatible with the update.py script.

schrein: I've tested the Pycom `update.py` script with the different shields (EB, Pysense, Pytrack) used on IoT-LAB. Good news, they work finally all ! So I will update the PR with removing the if statement for the EB shield; and by default use the Pycom scripts for properly erase the filesystem during setup.

* expansion board 3.1 + LoPy4

```
$ python ./updater.py list
/dev/cu.usbmodemPyf25f471  [Expansion3 - Pycom] [USB VID:PID=04D8:EF98 SER=Pyf25f47 LOCATION=20-1.2.4]

python ./updater.py -p /dev/cu.usbmodemPyf25f471 erase_fs
Running in PIC mode
Flash file system erasure successful
```

* pysense v1.0 + LoPy4

```
$ python ./updater.py list
/dev/cu.usbmodemPy3434341  [Pysense - Pycom] [USB VID:PID=04D8:F012 SER=Py343434 LOCATION=20-1.2.4]


$ python ./updater.py -p /dev/cu.usbmodemPy3434341 erase_fs
Running in PIC mode
Flash file system erasure successful
```

* pytrack v1.0 + LoPy4

```
$ python ./updater.py list
/dev/cu.usbmodemPy3434341  [Pytrack - Pycom] [USB VID:PID=04D8:F013 SER=Py343434 LOCATION=20-1.2.4]

$ python ./updater.py -p /dev/cu.usbmodemPy3434341 erase_fs
Running in PIC mode
Flash file system erasure successful
```

gateway_code/open_nodes/node_pycom.py
```diff
@@ -37,13 +39,28 @@
...
-    b"os.mkfs('/flash')\r\n",  # Erase the flash
+    b"os.fsformat('/flash')\r\n",  # Erase the flash
 )
 PYCOM_RESET_SEQUENCE = (
     b"import machine\r\n",
     b"machine.reset()\r\n",
 )
 
+PYCOM_UPDATE_BIN = "sudo /usr/bin/python3 " \
+    + "/usr/local/share/pycom/eps32/tools/fw_updater/updater.py"
```
aabadie: The sudo command is not available on gateway images. The gateway instance is running with the www-data user with limited rights. Why do you need `sudo` ? Do you think it could work without ?

aabadie (approved):
Looks good to me, let's merge.

PR #143, opened 2021-12-20
add support for JLINK Segger toolchain and OpenMoteb

schrein:
* add  JLINK Segger toolchain node type (support only binary firmware, not elf)
* add OpenMoteb implementation


bin/rules.d/openmoteb.rules
```diff
@@ -0,0 +1,2 @@
+# open nodes openmoteb
+KERNEL=="ttyUSB1" , SUBSYSTEMS=="usb", ATTRS{idVendor}=="0403" , ATTRS{idProduct}=="6010" , MODE="0664", GROUP="dialout", SYMLINK+="iotlab/ttyON_OPENMOTEB"
```
aabadie: I think this is wrong and is copy-pasted from M3/Fox rules. How can this work ? Did you change the IDs in the programmer ?

aabadie (commented):
Some comments:

- [x] it's possible to use an elf, just convert it to bin before flashing. We did similar things in the past for edbg (converting from elf to hex)
- [x] there are a lot of style/lint issues reported by the CI
- [x] There's something strange with the udev rules, idProduct/idVendor fields are the same as for the M3/Fox but it's a different thing (a JLink Mini Edu or something I guess). I think it would be useful to document the setup somewhere, maybe in the opennode docstring.
- [ ] the segger programmer class has no unit test. I know they are dumb but are useful to check things behave as they should and ensure important lines are covered by tests
- [ ] I haven't checked in details but is the segger programmer integrated in the `iotlab_flash` cli ? That would be useful of an openmote is plugged behind an open Linux

bin/rules.d/openmoteb.rules
schrein: > Some comments:
> 
> * it's possible to use an elf, just convert it to bin before flashing. We did similar things in the past for edbg (converting from elf to hex)

Yes indeed, we can to that. So we will assume to flash the bin files at a default offset.

> * there are a lot of style/lint issues reported by the CI

As usual :-)

> * the segger programmer class has no unit test. I know they are dumb but are useful to check things behave as they should and ensure important lines are covered by tests

Ok need to improve this too.

> * I haven't checked in details but is the segger programmer integrated in the `iotlab_flash` cli ? That would be useful of an openmote is plugged behind an open Linux

No, this could be another improvement in another PR.

> * There's something strange with the udev rules, idProduct/idVendor fields are the same as for the M3/Fox but it's a different thing (a JLink Mini Edu or something I guess). I think it would be useful to document the setup somewhere, maybe in the opennode docstring.

Will check that.

Thanks for the comments.

bin/rules.d/openmoteb.rules
```diff
@@ -0,0 +1,2 @@
+# open nodes openmoteb
+SUBSYSTEM=="tty", SUBSYSTEMS=="usb", ATTRS{idVendor}=="0403" , ATTRS{idProduct}=="6010" , ENV{ID_USB_INTERFACE_NUM}=="01", MODE="0664", GROUP="dialout", SYMLINK+="iotlab/ttyON_OPENMOTEB"
```
aabadie, line 2: I think I understand why the product and vendor id are the same as Fox and nearly the same as M3 and A8-M3 open nodes/control node.
To access the serial port your setup uses an FTDI USB to serial converter. These IDs are FTDI IDs: https://www.ftdichip.com/Support/Documents/TechnicalNotes/TN_100_USB_VID-PID_Guidelines.pdf

That might be worth mentioning it here or somewhere in the open node documentation (in the class doc string for example) or in both places.

gateway_code/open_nodes/common/node_segger.py
```diff
@@ -0,0 +1,139 @@
...
+        LOGGER.info('segger node debugger start')
+        return self.segger.debug_start()
+
+    def debug_stop(self):
+        """ Stop segger node debugger """
+        LOGGER.info('segger node debugger stop')
+        return self.segger.debug_stop()
+
+    @staticmethod
+    def status():  # pylint: disable=arguments-differ
```
aabadie: Instead of silencing, just use the right interface:

```suggestion
    def status(self):
```

gateway_code/open_nodes/common/node_segger.py
```diff
@@ -0,0 +1,139 @@
...
+        """ Check segger node status """
+        # Status is called when open node is not powered
+        return 0
+
+    @classmethod
+    def verify(cls):  # pylint:disable=unused-argument
+        """ This class use binaries firmware """
+        return 0
```
aabadie: This is not needed anymore since the open node now also supports elf files.

gateway_code/open_nodes/node_openmoteb.py
```diff
@@ -0,0 +1,40 @@
...
+    """ Open node Openmoteb implementation """
+
+    TYPE = 'openmoteb'
+    FW_IDLE = static_path('openmote-b_idle.elf')
+    FW_AUTOTEST = static_path('openmote-b_autotest.elf')
+    JLINK_DEVICE = "CC2538SF53"
+    JLINK_IF = "JTAG"
+    JLINK_RESET_FILE = static_path("openmoteb_reset.seg")
+    # CC2538 default flash_addr 0x200000
+    JLINK_FLASH_ADDR = 2097152
```
aabadie: You can store the flash address as hex, no need to  convert it to base 10.

```suggestion
    # CC2538 default flash_addr 0x200000
    JLINK_FLASH_ADDR = 0x200000
```

aabadie (commented):
There are still a few things to improve I think. The unittest is also still missing.

gateway_code/open_nodes/common/node_segger.py
```diff
@@ -0,0 +1,133 @@
...
+
+from gateway_code.utils.segger import Segger
+from gateway_code.utils.serial_redirection import SerialRedirection
+
+LOGGER = logging.getLogger('gateway_code')
+
+
+class NodeSeggerBase(OpenNodeBase):
+    # pylint:disable=no-member
+    """ Open node JLink Segger implemention """
```
aabadie:
```suggestion
    """ Open node JLink Segger implementation """
```

PR #147, opened 2022-04-06
gateway_code: fix recents lint reported

aabadie:


aabadie:
All lints are fixed. Just waiting for integration tests to complete locally.

PR #148, opened 2022-04-21
open_nodes: add support for stm32f746g-disco and stm32f7508-dk

aabadie:
This PR adds support for a couple of ST development kit that will be useful for the 5G mMTC project.


aabadie:
Tested the support on devsaclay and it works on stm32f7580-dk.

PR #149, opened 2022-04-21
gateway_code/static: update idle and autotest firmware

aabadie:
This PR updates the idle and autotest firmwares with new versions generated from [ci-firmwares](https://github.com/iot-lab/ci-firmwares).

The main changes in the firmwares are:
- migrate the autotest to ztimer (instead of nearly deprecated xtimer)
- clear eeprom memory on board that provide one (b-l072z-lrwan1)


aabadie:
be careful, this is an important change, especially for the integration tests (even if the changes in the firmware are quite minimal).

